PR #6, opened 2023-02-08
Initial P1P support

AdrianGarside:
Changes:
- Added optional serial number field to config flow. If provided it is passed into the initial connection attempt and used to subscribe to the specific mqtt topic that the P1P requires you only subscribe to.
- Removed duplication in initial model creation so there's only one update code path.
- Added exception handling around returned data from the mqtt topic - sometimes it's incomplete.
- Logged the binary returned data so we can see it even when we hit that case and can't parse to json (it's just a string so is equally understandable).
- Added an mqtt disconnect handler so we can see if/when the mqtt connection gets dropped unexpectedly along with the reason code.
- Renamed update_from_dict methods to just update since it seemed unnecessarily wordy
- Added more logging for my own understanding.

I'm brand new to python so bear with me :). I'm also happy to split these changes apart to take them in pieces.

I should probably get the lights working again since I disabled those as I was getting things up and running with my P1P.

### DIFF
--- a/custom_components/bambu_lab/config_flow.py
+++ b/custom_components/bambu_lab/config_flow.py
@@ -27,7 +27,8 @@ class BambuLabFlowHandler(ConfigFlow, domain=DOMAIN):
         if user_input is not None:
             bambu = BambuClient(user_input[CONF_HOST])
             LOGGER.debug("Config Flow: Trying Connection")
-            serial_number = await bambu.try_connection()
+            serial = user_input["serial"]
+            serial_number = await bambu.try_connection(serial)
 
             if serial_number:
                 return self.async_create_entry(
@@ -43,6 +44,6 @@ class BambuLabFlowHandler(ConfigFlow, domain=DOMAIN):
 
         return self.async_show_form(
             step_id="user",
-            data_schema=vol.Schema({vol.Required(CONF_HOST): str, vol.Required("name"): str}),
+            data_schema=vol.Schema({vol.Required(CONF_HOST): str, vol.Required("name"): str, vol.Optional("serial"): str}),
             errors=errors or {},
         )
--- a/custom_components/bambu_lab/coordinator.py
+++ b/custom_components/bambu_lab/coordinator.py
@@ -44,7 +44,7 @@ class BambuDataUpdateCoordinator(DataUpdateCoordinator):
             LOGGER.debug("Use MQTT: Listen")
             self.client = BambuClient(self._entry.data[CONF_HOST])
             self.client.connect(callback=message_handler)
-            self.client.subscribe(self._entry.data['serial'])
+            #self.client.subscribe(self._entry.data['serial'])
 
         # async def listen() -> None:
         #     def on_message(client, userdata, message):
--- a/custom_components/bambu_lab/pybambu/bambu_client.py
+++ b/custom_components/bambu_lab/pybambu/bambu_client.py
@@ -17,10 +17,10 @@ class BambuClient:
     def __init__(self, host: str):
         self.host = host
         self.client = mqtt.Client()
-        self._serial = None
+        self._serial = ""
         self._connected = False
         self._callback = None
-        self._device: Device | None = None
+        self._device = Device()
 
     @property
     def connected(self):
@@ -32,6 +32,7 @@ class BambuClient:
         """Connect to the MQTT Broker"""
         self._callback = callback
         self.client.on_connect = self.on_connect
+        self.client.on_disconnect = self.on_disconnect
         self.client.on_message = self.on_message
         LOGGER.debug("Connect: Attempting Connection")
         self.client.connect(self.host, 1883)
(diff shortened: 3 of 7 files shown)


AdrianGarside:
Looks like my most recent cleanup changes to try and get this ready for a pull request broke the config flow and I must have messed up my testing of that.

WolfwithSword:
Given the new MQTT changes, should probably look towards having the Lan Access Code also be in the config flow for the MQTT password (not sure if P1P has this requirement yet but it wouldn't hurt). Doesn't need to be this PR unless it affects the P1P too.

Not sure of other's opinions but I think it would be best to make the Serial num a strict requirement of the flow so we never have to rely on the X1C's being auto picked up unlike the P1P's. Then we can have an explicit dropdown in the config for printer type (X1C, X1, P1P), which can be later used for sensor naming, model information, and for defining other flows (such as the mqtt creds if necessary, some sensors not available otherwise etc). This makes the flow more standardized between printers, and each printer config can use what's provided if/when necessary.

AdrianGarside:
Yes, I was thinking the same. Part of the reason I was rushing this PR a bit was to get it out of the way so I could work on those changes. I think I can code up the MQTT authentication changes but I won't actually be able to test them given I only have a P1P which hasn't been updated with that change yet. I'm sure it's coming though so I was planning to just make access code required from the get go.

I think it should be possible to identify the printer type automatically so no need to ask the user to tell the integration which avoids the potential for a mismatch. That was next on my list since the integration is currently hardcoded to claim the printer is an X1C and I'd like to fix that. 

Always requiring the serial isn't a big deal IMO since even if it's not strictly necessary for the X1, this should be a one-time step so a few extra seconds there isn't significant in the longer term and does help create a clearer configuration experience by removing choice.

WolfwithSword:
Sounds great to me, only thing is re: the choice of printer - P1P and X1C may be easy to tell apart but I'm not sure about X1 vs X1C. Mainly thought it would just be a easy/convenience thing to make it a selection.

Then again we don't want to bloat the config, still debating if it would be a good idea to have a text config field just to "name" the printer if you want to have it easier to identify/match up to a name from the app/slicer or other.

Another benefit to making the printer model a selection, is that you could probably configure certain sensors/entities to not even be created based on that config, or just behave differently. One example would be when we eventually add in the speed selection, it wouldn't be added to the P1P, and then the external spool config (currently) is only existing for the P1P and not X1C etc.

AdrianGarside:
I also want to work out if there's a way I can explicitly trigger that massive json dump of all state so that at startup we can populate all existing sensor state accurately. So far you've shown me how to get the version data on demand so I'm hopeful there's a similar command that will generate the large state dump.

AdrianGarside:
> Then again we don't want to bloat the config, still debating if it would be a good idea to have a text config field just to "name" the printer if you want to have it easier to identify/match up to a name from the app/slicer or other.

This got me wondering if the name is exposed somewhere in the json payloads. Be nice to self-discover the name the user already gave their printer instead of asking for it. Currently the passed in name isn't used - there's a hard coded value of 'change' with a TODO that needs to be addressed by plumbing that passed in name through and/or auto-discovering and passing that around.

> Another benefit to making the printer model a selection, is that you could probably configure certain sensors/entities to not even be created based on that config, or just behave differently. One example would be when we eventually add in the speed selection, it wouldn't be added to the P1P, and then the external spool config (currently) is only existing for the P1P and not X1C etc.

P1P has speed selection. But yes, it may be useful to not have sensors exposed to home assistant that don't exist. Just not sure it's a big deal in practice given we can flag the sensors as unavailable. So they're present as entities but flagged as not available.

WolfwithSword:
> This got me wondering if the name is exposed somewhere in the json payloads.

Unfortunately not, it's entirely configurable but not exposed like that. Ironically, I tried testing different "get" commands to info/system/print. In "info", any unrecognized command returns the get_version info which is interesting. I tried sending command "push_status" and "get_status" to "print"... push_status returned an error so maybe I'm missing config, but anything else didn't work. It must need some kind of param or other info.

> So they're present as entities but flagged as not available.

That's my workplace mind thinking "some end user is going to complain" 😆

AdrianGarside:
> That's my workplace mind thinking "some end user is going to complain" 😆

True. But bigger fish to fry right now :)

We need a name to create unique entities if you connect multiple printers. We could just use the serial number but that's kinda gross but probably quicker to do for now.

WolfwithSword:
I think serial number is probably fine for the unique id unless we provide a name. Either way, would need some sort of unique identifiers to at the very least, use the native connected-to device linking for AMS devices - which also have their own unique serial numbers and id's.

AdrianGarside:
Folks will be able to give things any friendly name they like in their dashboard so serial number as the unique part of the entity names is probably fine and then we can drop the name field in the config flow.

AdrianGarside:
Made the changes to require the lan code and serial number and eliminate asking for a name. Extended the try connection timeout to 10s as I was seeing 5s suddenly wasn't long enough to guarantee seeing a message.

AdrianGarside:
@greghesp any concerns with this PR? I'd like to complete it to have a bash at adding the TLS support.

WolfwithSword:
> removing the autodiscovery of the serial

That was my recommendation so we could provide a consistent user-config flow regardless of printer. We're going to have to configure parts regardless (ip/host, access code pw, etc) may as well toss that in there., especially if autodiscovery one day stops working and since it's info we know beforehand etc.

If we keep the option in then I'd recommend a multi-stage flow where you select your printer model first from a dropdown (I've tried but to no avail can I tell the printers apart via their json, at best if you parse their responses from get_version there should be hardware_version differences, but that requires a connection first). That way if P1P you need to enter it on another screen, and X1/X1C you don't. This would be beneficial in the meantime since P1P doesn't have the MQTT credential stuff yet too, but it should be soon.

An alternative would be to during the flow, keep the user locked as it "attempts" to get the serial number, and if it fails to get one after a 10 second timeout, force them to provide it?

PR #163, opened 2023-07-20
Update README.md with revised P1S support

ianschmitz:
The P1S includes the aux cooling fan by default. It also includes a chamber light.

image
image

Let me know if you'd like to check anything else 😃


AdrianGarside (approved):

AdrianGarside:
Thanks for catching and fixing the mistakes in the readme.

Issue #93, opened 2023-04-14
[Bug] Filament lists is out of date and incomplete


AdrianGarside:
### Describe the bug

I found a few missing ones by chance. But reviewing the bambu studio source there's a lot of new filaments we need to add to the FILAMENT_NAMES list

### To Reproduce

.

### Expected Behaviour

.

### What device are you using?

P1P

### Firmware Version

Latest

### Home Assistant Version

2023.3.6

### Component Version

Latest

renssies:
Can confirm, "Bambu PLA Marble" is missing

PR #168, opened 2023-07-30
Updated filaments

fmeus:
Updated and sorted the list of filaments


fmeus:
Cloud be solution to  #93

PR #178, opened 2023-08-24
Add entity translation support

piitaya:
This PR add translation support to any entity following [Home Assistant documentation](https://developers.home-assistant.io/docs/internationalization/core#entities).

### Translations

`strings.json` has been deleted because it's not needed for custom components. Only `en.json` is useful.
Translation keys has been added to any entities.

I also added french translations as it's my native language.

### Entity options

`options` has been added to current stage, speed profile and printing speed to add autocompletion in automation editor and translations.

### Lowercasing string value

`Unknown` has been changed to  `unknown` so it can be translated too.
Speed value has be converted to lowercase to support translations too.

### Device class for online entity

Online entity has been changed to running instead of power because power is for electricity power detection not device power on.

Feel free to give any feedback if you find something strange or wrong.


custom_components/bambu_lab/fan.py
```diff
@@ -32,17 +32,17 @@ class BambuLabFanEntityDescription(FanEntityDescription, BambuLabFanEntityDescri
 FANS: tuple[FanEntityDescription, ...] = (
     BambuLabFanEntityDescription(
         key="cooling_fan",
-        name="Cooling Fan",
-        value_fn=lambda device: device.fans.cooling_fan_speed
+        translation_key="cooling_fan",
+        value_fn=lambda device: device.fans.cooling_fan_speed,
```
AdrianGarside, line 36: I think, but am not sure, that this extra comma will cause problems? But maybe not if this is running fine for you as-is. But it is inconsistent with the other code in this file.

AdrianGarside (commented):
What do we need to do as/when we add new translatable strings - do we need to duplicate them (untranslated / online translated) into the fr.json file or is it ok to just add them to the en.json file and it'll automatically use them until someone (like yourself) adds the missing translations?

Issue #173, opened 2023-08-09
[Feature] Please add camera support (now working on X1C)


Yoghoo:
### Describe the feature

Since the last firmware it's possible to view the camera of the X1C in HA.

For instructions see: https://www.wolfwithsword.com/bambulabs-x1c-camera-in-home-assistant/

### What device is this for?

X1C

WolfwithSword:
One thing to note when this gets added (which would have the benefit of having the camera be linked to the device) is that for it to work, the "Lan Liveview" mode has to be turned on via the printer for the X1C, so it should be added as a prerequisite if the user would want to see the camera. LAN Mode is not required.

In order to tell if it is enabled or not, and likely best approach for whether or not the device should setup the camera entity is to watch out for the ipcam messages.

In the `print` payload, if the `ipcam` exists with `rtsp_url` inside, if the value is `disable` then the lan liveview is not enabled. Otherwise, when enabled, it starts with `rtsps` and will actually provide the full url (without authentication or port, both of which not necessary for HA generic camera configuration).

So likely in this case, the flow could be that once the integration receives the ipcam payload, if the rtsp_url is filled with disable, no camera entity exists or mark it as unavailable. Otherwise, it directly gives the URL with correct IP each time, and since the integration already gets the access code, it should be perfect to use and should only be disabled again if the value turns to `disable` or printer goes offline.

Do note that it can be unstable on fresh boot after firmware updates, or sometimes if the printer was booted initially with the setting off. Restarting the printer may fix it. I also have found that sometimes the generic camera integration needs reloading if it is stuck on unavailable for some time - may not be needed for this integration though, would need testing (only encountered it once in the past 2w with manual generic camera).

bvrielink:
> Do note that it can be unstable on fresh boot after firmware updates, or sometimes if the printer was booted initially with the setting off. Restarting the printer may fix it. 

Had this problem as well. Restarting it didn't help (tried a couple of times). I then enabled LAN Mode and directly disabled it again. After another restart it finally worked. Thanks for the article btw. :)

tretabyte:
I'm having the same issue as it everything seems to be turned on, sometimes I can see something for like half a second but then I only see 'Internal Error' or 'connection refused' which is odd, maybe turn it off,on,off, on with a few restarts in between might solve the issue.
since we are talking about cameras, is there a way to pull the print preview that shows the rendered object being printed? it would be great if we could have access to that preview.

WolfwithSword:
[...]

PR #184, opened 2023-09-05
Add Camera functionality

greghesp:
Added Camera functionality as per https://github.com/greghesp/ha-bambulab/issues/173

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ If AMS(s) are present, an additional 'Current Tray' sensor is present on the Pri
 
 ### Lights
 
-| Sensor        | X1C                | X1                 | P1P                | P1S                |  
+| Light         | X1C                | X1                 | P1P                | P1S                |  
 |---------------|--------------------|--------------------|--------------------|--------------------|
 | Chamber Light | :white_check_mark: | :white_check_mark: | :heavy_check_mark: | :white_check_mark: |
 
@@ -56,7 +56,7 @@ If AMS(s) are present, an additional 'Current Tray' sensor is present on the Pri
 
 This currently exposes the following Buttons:
 
-| Sensor | X1C                | X1                 | P1P                | P1S                |
+| Button | X1C                | X1                 | P1P                | P1S                |
 |--------|--------------------|--------------------|--------------------|--------------------|
 | Pause  | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: |
 | Resume | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: |


custom_components/bambu_lab/camera.py
```diff
@@ -0,0 +1,80 @@
...
+    _attr_brand = "Bambu Lab"
+
+    def __init__(
+            self,
+            coordinator: BambuDataUpdateCoordinator,
+            config_entry: ConfigEntry
+    ) -> None:
+        """Initialize the sensor."""
+
+        self._attr_unique_id = f"{config_entry.data['serial']}_camera"
```
AdrianGarside, line 42: For consistency with other devices should we prepend this with the model name?

custom_components/bambu_lab/pybambu/models.py
```diff
@@ -285,13 +320,15 @@ def print_update(self, data):
         self.remaining_time = data.get("mc_remaining_time", self.remaining_time)
         if data.get("gcode_start_time") is not None:
             self.start_time = start_time(int(data.get("gcode_start_time")))
+        # TODO: Check this logic actually works
+        if self.gcode_state is not 'Running':
```
AdrianGarside, line 324: We might have some case inconsistency issues lurking here for this variable. I see for the localization changes this is expected to be lower case (maybe, not sure it does a case sensitive match) bit the actual mqtt content is all upper case. And we don't lower case it when we parse/save the string.

custom_components/bambu_lab/translations/en.json
```diff
@@ -360,6 +360,11 @@
           }
         }
       }
+    },
+    "camera": {
+      "camera": {
+        "name": "Camera"
```
AdrianGarside, line 366: If we don't add this to the fr-fr file, does it just use the English one instead? I.e. Do we have to add something to all the non-English json files or can localization come later (from native speakers)?

custom_components/bambu_lab/pybambu/models.py
```diff
@@ -125,6 +130,38 @@ def TurnChamberLightOff(self):
...
+        self.timelapse = ''
+
+    def print_update(self, data):
+        """Update from dict"""
+        # "ipcam": {
+        #   "ipcam_dev": "1",
+        #   "ipcam_record": "enable",
+        #   "mode_bits": 2,
+        #   "resolution": "1080p",
+        #   "rtsp_url": "rtsps://192.168.1.64/streaming/live/1",
```
AdrianGarside, line 154: Not something to worry about for this PR but we might be able to populate the camera availability based on whether this exists in the mqtt payload instead of gating it on the model. But probably more work than benefit.

custom_components/bambu_lab/camera.py
AdrianGarside: Never mind we're already inconsistent and this is unique ID so the model number is irrelevant. Good to go on this.

AdrianGarside (approved):

PR #190, opened 2023-09-09
Update french translations and add empty attribute translation

piitaya:
Empty translations was missing in spool attributes.
I also updated the french translations.


AdrianGarside:
@piitaya if we forget to add a translation again (but English is present) does HA use the English one?

piitaya:
@AdrianGarside Yes! No need to add all languages 🙂

PR #200, opened 2023-09-14
Fixed some typos and correction to string comparison

fmeus:
Textual corrections in HMS message and stage translations. Also correction to string comparison on gcode_state


AdrianGarside:
Thanks.

Issue #195, opened 2023-09-11
[Bug] When printer turn off and the turn on, the entities integration don't update the status


josemabcn:
### Describe the bug

I have a remote switch connected to X1C. The integration is working well, but if I turn OFF the printer, and then turn ON, all the integration entities show the last state before turn off. The light switch for example don't works, I turn the light switch OFF but the light continues ON, etc. I need to go to the integration and reload the integration and then all the entities works well. But If I turn the printer off another time the same problema.

### To Reproduce

Turn off the printer

### Expected Behaviour

When the printer turn ON all the entities needs to be available and working.

### What device are you using?

X1C

### Firmware Version

01.06.03.00

### Home Assistant Version

Supervisor 2023.08.3 / Operating System 10.5 / Frontend 20230908.0 - latest

### Component Version

V1.4.9

AdrianGarside:
That's odd. The logic has been there to detect the printer turning off and then back on again while the integration is running for quite a while. I'll need debug logs from the integration to see if I can work out why this isn't working - there's been no recent changes in this area and it's still working correctly with my P1P.

josemabcn:
I dont' know exactly how to download the depuration log, in the register logs I have the following errors when I try to activate the fans and lights and they don't work:

`ValueError: Sensor sensor.x1c_00m09a360200242_print_status provides state value '', which is not in the list of options provided
2023-09-12 12:13:23.119 ERROR (SyncWorker_10) [custom_components.daily.sensor] unable to convert unavailable to float. Please check the source sensor (sensor.velocidadviento) is available.
2023-09-12 12:13:23.121 ERROR (SyncWorker_10) [custom_components.daily.sensor] unable to convert to float. Please check the source sensor (sensor.velocidadviento) is available.
2023-09-12 12:13:23.241 ERROR (SyncWorker_14) [custom_components.daily.sensor] unable to convert unavailable to float. Please check the source sensor (sensor.velocidadviento) is available.
2023-09-12 12:13:23.242 ERROR (SyncWorker_14) [custom_components.daily.sensor] unable to convert to float. Please check the source sensor (sensor.velocidadviento) is available.
2023-09-12 12:14:07.418 DEBUG (MainThread) [custom_components.bambu_lab] HA POLL: MQTT connected: True
2023-09-12 12:14:07.418 DEBUG (MainThread) [custom_components.bambu_lab] Finished fetching bambu_lab data in 0.000 seconds (success: True)
[...]

PR #202, opened 2023-09-14
Add pause print status

piitaya:
According to https://github.com/greghesp/ha-bambulab/issues/195, pause is a valid print status. I reproduced it on my P1S.


AdrianGarside:
@piitaya a couple of nit comments.

AdrianGarside:
Also we need to pre-emptively add 'SLICING' as an option so we don't start having issues when that feature rolls out.

piitaya:
I added slicing and init state to prevent issue. I also update a french translation.

PR #225, opened 2023-09-22
Fix start time sensor. Adjust end time sensor.

AdrianGarside:
Previously we set the (generated) end time sensor to unavailable when remaining time reaches zero. But really we want to persist it since it's still valid after the print ends. That means we need to not wipe it out when remaining time reaches zero. So now we regenerate the end time date whenever remaining time changes which means we set it one last time when remaining time reaches zero vs making it unavailable. End time will start unavailable and remain so until the first print occurs - then it's always going to show the last print or the in progress print estimated end time.

Start time needs the same handling. It will start unavailable on printer startup and then get set on print start and be persistent until the next print start. So in most use, only a printer restart will make the start time sensor go unavailable. And end time I think will survive even that - only an integration reload will set that back to unavailable.


AdrianGarside:
Hmm... this will have the not wholly desirable behavior that we'll set end time to 'now' on integration restart after printer restart.

PR #230, opened 2023-09-23
Generate start time for P1P/S

AdrianGarside:
These don't generate the start_time mqtt payload natively. But we can simulate it by detecting remaining time changing from 0 to some value. The main limitation is that if the integration is started/restarted mid-print it won't be able to generate a correct value so the start time will remain unavailable in that case.


AdrianGarside:
Ooh, bambu handy app got updated on me with the new maker world stuff.

PR #258, opened 2023-10-23
Fix state comparisons for blueprint

claytono:
This changes the states in the automation blueprint to look at the actual states exposed instead of the localized name.  Without this the automations didn't actually work for me.  I'm not sure if this is something odd with my home assistant config or not, but the strings being compared here are the localized versions, so I'm guessing this might be needed in order to not be localization specific?

### DIFF
--- a/blueprints/wled_controller.yaml
+++ b/blueprints/wled_controller.yaml
@@ -57,19 +57,19 @@ action:
         conditions:
           - condition: state
             entity_id: !input printer_stage
-            state: Scanning Bed Surface
+            state: scanning_bed_surface
             alias: Scanning Bed Surface
           - condition: state
             entity_id: !input printer_stage
-            state: Cleaning Nozzle Tip
+            state: cleaning_nozzle_tip
             alias: Cleaning Nozzle Tip
           - condition: state
             entity_id: !input printer_stage
-            state: Calibrating Extrusion
+            state: calibrating_extrusion
             alias: Calibrating Extrusion
           - condition: state
             entity_id: !input printer_status
-            state: Offline
+            state: offline
             alias: Printer is off
         alias: And Lidar is NOT on
     then:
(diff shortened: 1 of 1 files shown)


AdrianGarside (approved):

filipkotian:
Hi @AdrianGarside I have just installed latest version, but these states are still not reporting correctly, I mean states like: scanning_bed_surface, inspecting_first_layer are not reported correctly so I can not turn off LED strip when Lida is working. Could you please have a look at it? many thanks!

AdrianGarside:
@filipkotian can you create a new issue for this please.

AdrianGarside:
@claytono are you seeing the same issue with the latest version?

filipkotian:
currently I see only this states: heatbed_preheating, cleaning_nozzle_tip, auto_bed_leveling, calibrating_extrusion, than just printing and unknown. All other states are not reported correctly for me. X1C

image

AdrianGarside:
I'm only seeing these on my P1P (unless they're so short the HA history UI won't show them):
image
idle, heatbed_preheating, cleaning_nozzle_tip, homing_toolhead, printing

I did notice that in the more recent firmware they look to have optimized print startup somewhat so that some of the steps are happening a bit more in parallel. And I think I recall seeing more states showing in my HA dashboard than just those.

Are you seeing the other states showing in Bambu Studio or the Handy app?

filipkotian:
Yes same here, in handy app I see some more like this one: 
IMG_1990

AdrianGarside:
I closely monitored the mqtt payload during a print start and we are correctly showing the current_stage. I see mc_print_sub_stage which appears (need to confirm) to track the loading of the filament. I think the output you see above is an HMS 'error' -     "0C00_0300_0003_000B": "Inspecting first layer",

They no longer appear to be using most of the values that stg_cur used to enumerate.

PR #272, opened 2023-11-19
Added description for HMS Error Code 0700_2x00_NNNN_NNNN

honcheng:
HMS Error code 0700_2x00_NNNN_NNNN where x is 1, 2, and 3 are not documented on Bambu Lab wiki. They showed up as unknown in HA. 

What's documented is 0700_0200_0003_0002 where it is when filament in tray 1 ran out. 
I've personally encountered the HSM Error code when Tray 3 and Tray 4 ran out, and the error code looks like these: 
0700_0220_0003_0002 and 0700_0230_0003_0002, so I think it is right to assume that the x is the index of the tray. 

These are 2 other similar error code that I assumed also changes with tray index, so I went ahead and added those too.


honcheng:
Here's the error code when tray 4 filament ran out. 
IMG_F12763FA4421-1

WolfwithSword:
There's a lot of HMS codes without wiki pages still.

The full (english) list can be found here: https://e.bambulab.com/query.php?lang=en so there's likely plenty more that could get added. Unfortunately bambu changed some stuff so it doesn't work with GET requests anymore (no scraping / bots / scripts / automations) but it could be manually copied/converted over.

List has all the codes, partial codes, and their description. Unfortunately it doesn't give information like severity level or module type, for that there's other bitwise operations that need to be done. Really wish it were just in the json file :(

PR #282, opened 2023-11-24
add Simplified Chinese localization

L1cardo:
Language tags: https://github.com/home-assistant/frontend/blob/dev/src/translations/translationMetadata.json

image


AdrianGarside:
Awesome. Thank you!

Issue #249, opened 2023-10-09
[Feature] Remaining % of filament in ams?


BobFish:
### Describe the feature

Seeing as the app and studio report this, is it possible to get the amount of filament remaining in each slot of the AMS reported in HA? Haven't seen this if it is already there but doesn't look like it.

Would be very useful for dashboards and automations if possible!

### What device is this for?

AMS

AdrianGarside:
@BobFish I don't have any bambu filaments currently. Can you capture a log file including use of one so I can see what the mqtt payload looks like?

AdrianGarside:
My guess it it'll be this:
          "tray": [
            {
              "id": "0",
              "remain": -1,

But with some value other than -1.  Maybe just gram value estimate?

WolfwithSword:
It'll require both a bambu spool and the filament amount/remaining update setting to be enabled (by default, writing to the remaining amount is not enabled I believe).

If you have the remaining tracking disabled, *all* spools will have a remaining of 0, regardless if Bambu or not.

With it enabled, for spools that don't have the bambu tag, it'll always have the "remain" value of -1, and you can determine if it's a bambu spool or not through various means also found in the "tray" object (tray_uuid won't be all 0's, for ex).

Could probably add it to the tray attributes, but to not be misleading, would need also show if it's one that has remaining-tracking enabled (so any bambu spool - not pulled from the info_idx or profile as some people may use bambu profiles to create custom ones. Can either treat -1 as not supported, or find out via the uuid's.). You'd also want to take care of the case for 0, since that may indicate it's not enabled.

Also - the remaining amount uses how much the spool rotates and the number of metres that go through the feeder back and forth to determine the amount remaining. It then uses this and the known information about filament density and diameter to determine the estimated value for tray_weight. There are no weight-sensors, it's purely calculated.

The actual value will be a percentage from 0-100. For example, I have still an old spool of the pla support that came with the printer over a year ago, came in a 500g amount on a full spool. I've used approximately half of it I'd wager. When I load it up, and in bambustudio enable the calculate remaining option (I've had it disabled since forever since I rarely use bambu filaments, and usually never in the AMS) it shows a value of `23`, and "tray_weight" is 250. 

Images after enabling the setting*

![image](https://github.com/greghesp/ha-bambulab/assets/12175651/b9d1b9f7-a8a4-41a2-9c10-32c9eae6ba43)
[...]

PR #284, opened 2023-11-24
Added remaining filament amount to AMS trays.

myMartek:
To solve the open issue #249 I created a pull request which implements the reading of the filament. My solution is quite simple: I simply added the remain parameter as it is. For filaments that are not Bambulabs or when the reading is disabled in the printers settings the value will simply be -1 and otherwise represent a percentage from 0 - 100%


myMartek:
I added a pull request which simply adds the remain value as it is to each AMS Slot. It can be -1 if there is no bambu filament or when the option to track the state is disabled but for me it is still a valid reading we can show as it is.

Bildschirmfoto 2023-11-24 um 22 28 08

myMartek:
I also added the tag_uid parameter for identifying the filament spool.

AdrianGarside:
main is our release branch - can you change this PR to be against develop instead please.

myMartek:
Sure!

AdrianGarside:
https://github.com/greghesp/ha-bambulab/releases/tag/v1.9.2-dev1

PR #299, opened 2023-11-30
Reapply camera URL fix

AdrianGarside:
The code changed slightly since the original fix and re-enabling cloud mqtt mode for connection means we don't necessarily have the right host name/IP to use so need to protect against that as well.

### DIFF
--- a/custom_components/bambu_lab/camera.py
+++ b/custom_components/bambu_lab/camera.py
@@ -68,8 +68,8 @@ class BambuLabCamera(BambuLabEntity, Camera):
 
             parsed_url = urlparse(self.coordinator.get_model().camera.rtsp_url)
             if self.coordinator.get_model().info.mqtt_mode == "local":
-                # For unknown reasons the returns rtsp URL sometimes has a completely incorrect IP address in it for the host.
-                # Since we know the correct IP (in local mqtt connection mode), rewrite the URL to have that.
+                # For unknown reasons the returned rtsp URL sometimes has a completely incorrect IP address in it for the host.
+                # Since we know the correct IP (but only in local mqtt connection mode), rewrite the URL to have that.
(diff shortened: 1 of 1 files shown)


custom_components/bambu_lab/camera.py
```diff
@@ -67,7 +67,12 @@ async def stream_source(self) -> str | None:
             # rtsps://192.168.1.1/streaming/live/1
 
             parsed_url = urlparse(self.coordinator.get_model().camera.rtsp_url)
-            url = fr"{parsed_url.scheme}://bblp:{self._access_code}@{parsed_url.netloc}{parsed_url.path}"
+            if self.coordinator.get_model().info.mqtt_mode == "local":
+                # For unknown reasons the returned rtsp URL sometimes has a completely incorrect IP address in it for the host.
+                # Since we know the correct IP (but only in local mqtt connection mode), rewrite the URL to have that.
+                url = fr"{parsed_url.scheme}://bblp:{self._access_code}@{self._host}/{parsed_url.path}"
```
godely, line 73: Please remove the backslash before {parsed_url.path} as that variable is already prefixed with a backslash. Fixed line should be:
`url = fr"{parsed_url.scheme}://bblp:{self._access_code}@{self._host}{parsed_url.path}"`

godely (commented):
URL is malformed due to double backslashes:
```
Error from stream worker: Error opening stream (ERRORTYPE_5, I/O error) rtsps://****:****@192.168.0.134//streaming/live/1
```

custom_components/bambu_lab/camera.py
AdrianGarside: #306

AdrianGarside: Fixed v1.9.3-dev2 release.

AdrianGarside: In my testing though that doesn't break the camera entity - the error above seems likely to just be bugs in the firmware.

PR #341, opened 2023-12-23
Simplify start and end time sensor values when they are for today.

AdrianGarside:


markfrancisonly:
All of my clocks show 12 hour time displays, and I won't have it any other way. 

It's much more difficult to parse a string that may be a "time" or maybe a "date time" string. So I feel this change is a step in the wrong direction, because the sensor value is not semantically stable, and "decoding" of the sensor value has to be duplicated all over e..g automations, localized text, etc. 

When the date is embedded into the sensor, it's possible to format the time using a template sensor for locality. Requesting rollback.

PR #347, opened 2023-12-24
en.json: Swap "Lan mode" for "Lan Only Mode"

reedy:
Makes the text match what is displayed on the printer


reedy:
Looks like the README.md wants updating too. Could be done in a different PR

AdrianGarside (approved):

PR #376, opened 2023-12-30
Update de.json

maddingamer:
Fixed german translation for "stage" as it was the same as "current_layer".


AdrianGarside:
Thanks.

PR #393, opened 2024-01-04
Blueprint: Lights off while calibrating_extrusion_flow

frittes:
Adds lights off while flow rate calibration with lidar to the blueprint.

### DIFF
--- a/blueprints/wled_controller.yaml
+++ b/blueprints/wled_controller.yaml
@@ -68,6 +68,10 @@ action:
             state: calibrating_extrusion
             alias: Calibrating Extrusion
           - condition: state
+            entity_id: !input printer_stage
+            state: calibrating_extrusion_flow
+            alias: Calibrating Extrusion Flow
+          - condition: state
             entity_id: !input printer_status
             state: offline
             alias: Printer is off


Freekers:
Thanks, noticed just yesterday this was missing. Do I need to re-create the automation based on this blueprint? Or is it sufficient to update the blueprint itself, seeing that the automation links to the blueprint .yml? Thanks

frittes:
Only update the blueprint file.

Freekers:
@frittes Shouldn't we also turn off the lights while auto bed leveling? I've noticed the LIDAR is in use during auto bed leveling, yet the WLEDs flash blue during this stage.

frittes:
Is there a change of the state entity while using the lidar? Or is the state also auto_bed_leveling?

Freekers:
> Is there a change of the state entity while using the lidar? Or is the state also auto_bed_leveling?
> 
> EDIT: Answering myself, it isn´t a change while bed surface scanning. The problem is that some states not shown from the possible attributes. inspecting_first_layer and calibrating_micro_lidar or scanning_bed_surface for example. It´s not a problem of the integration, the bambu itself does also not have the states.
> 
> We can add a switch for first layer inspection to the blueprint, but then we have lights off while bed leveling and lights on while first layer inspection, because it´s while printing.
> 
> From my point of view it makes no sense.

Sorry for the delayed reply. Indeed, you're right, the state entity while using the lidar to inspect the first layer is 'Printing', so that's not very useful. 

However, I did some testing, while inspecting the first layer the 'Nozzle Target Temperature' drops to 140 and 'Current Layer' is equal to 2. Perhaps we can create a condition based on a combination of these two states?

frittes:
Tried it, but the nozzle target temp and the current layer is not updated fast enough by the integration. Tried with a little object, but also on bigger objects the conditions depends on the update interval of the integration.

AdrianGarside:
We get updates from the printer about once a second and don’t spend much time processing them. So other that extra latency from the update cadence we present changes as fast as the printer tells us about them.

frittes:
Ok, then the printer does not send the target temp and current layer while first layer inspection. In the Slicer they are visible.

AdrianGarside:
I'm pretty sure it does since I'm sure I see the nozzle temp drop to 140 while it's doing the inspection in home assistant.

PR #409, opened 2024-01-08
Wiki link incorrect for non AMS related errors

fmeus:
HMS errors not related to the AMS are being regarded as being AMS related and thus were made generic.

Now the same method is followed as in get_HMS_error_text (check if error is part of HMS_AMS_ERRORS).


AdrianGarside:
Thanks. I had noticed I'd caused this but hadn't got around to seeing what I messed up.

PR #513, opened 2024-04-23
add filament name, bambu pla galaxy and custom filament aliz pteg

meishild:
add filament name

GFA15: bambu pla galaxy
P42b3ffe: custom filament aliz pteg

Will customization filament be available in the future?


custom_components/bambu_lab/pybambu/const.py
```diff
@@ -162,6 +163,7 @@ class FansEnum(Enum):
     "GFT97": "Generic PPS",
     "GFT98": "Generic PPS-CF",
     "GFU00": "Bambu TPU 95A HF",
+    "P42b3ffe": "ALIZ PETG"
```
AdrianGarside, line 166: Custom filaments are unique to you so this can't be added to the integration.

Issue #573, opened 2024-08-12
Logged error about fans


geofffarmer:
### Describe the bug

I have the following error in my system log:

Logger: homeassistant.components.fan
Source: components/fan/__init__.py:295
integration: Fan (documentation, issues)
First occurred: 16:21:32 (4 occurrences)
Last logged: 16:21:32

Entity None (<class 'custom_components.bambu_lab.fan.BambuLabFan'>) does not set FanEntityFeature.TURN_OFF but implements the turn_off method. Please create a bug report at https://github.com/greghesp/ha-bambulab/issues
Entity None (<class 'custom_components.bambu_lab.fan.BambuLabFan'>) does not set FanEntityFeature.TURN_ON but implements the turn_on method. Please create a bug report at https://github.com/greghesp/ha-bambulab/issues

### To Reproduce

Seems to be since update to HA Core 2024.8

### Expected Behaviour

No error

### What device are you using?

A1

### Diagnostic Output

```shell
Logger: homeassistant.components.fan
Source: components/fan/__init__.py:295
integration: Fan (documentation, issues)
First occurred: 16:21:32 (4 occurrences)
Last logged: 16:21:32

Entity None (<class 'custom_components.bambu_lab.fan.BambuLabFan'>) does not set FanEntityFeature.TURN_OFF but implements the turn_off method. Please create a bug report at https://github.com/greghesp/ha-bambulab/issues
Entity None (<class 'custom_components.bambu_lab.fan.BambuLabFan'>) does not set FanEntityFeature.TURN_ON but implements the turn_on method. Please create a bug report at https://github.com/greghesp/ha-bambulab/issues
```

Robbe-B:
FYI See => https://developers.home-assistant.io/blog/2024/07/19/fan-fanentityfeatures-turn-on_off/

AdrianGarside:
Fixed recenty

PR #583, opened 2024-08-26
Remove warnings about FanEntity for TURN_ON and TURN_OFF

DissiNL:
Fix:
(<class 'custom_components.bambu_lab.fan.BambuLabFan'>) does not set FanEntityFeature.TURN_ON but implements the turn_on method (<class 'custom_components.bambu_lab.fan.BambuLabFan'>) does not set FanEntityFeature.TURN_OFF but implements the turn_off method


DissiNL:
Fixes #573

nzrutman:
This fix works for me
👍

PR #625, opened 2024-10-23
Fix slicer_settings access denied

xz0609:
Add “User-Agent” field to the header to fix the issue of client.get() returning access denied (403).


AdrianGarside:
Awesome. I did wonder if it was a missing user agent but I was asleep so couldn't test it :)

PR #643, opened 2024-10-27
Merge develop -> main

AdrianGarside:


custom_components/bambu_lab/pybambu/models.py
```diff
@@ -1191,6 +1191,7 @@ class PrintErrorList:
     _count: int
 
     def __init__(self, client):
+        self._error = None
         self._client = client
         self._error = {}
```
Chreece, line 1196: should that still stay?

AdrianGarside: Good catch. No, that should be removed although it's probably benign as in practically all cases it'll be overwritten by None on the very first printer update.

PR #672, opened 2024-11-04
Added number compontents to control temperature of heatbed and nozzle

Skydev0h:
Added number components that allow to adjust heatbed and nozzle temperature using Number components.

For X1* printers in LAN mode and any printer in Cloud mode "instant set" gcode commands are used, while for P1*/A1* printers in LAN mode "set and wait" commands are used (because "instant set" commands are blocked).

The "set and wait" commands are quite finnicky and take some time to execute and update, therefore it is possible to hide the new components for situations where "set and wait" is required with slight modification of the code.

Feel free to make your own adjustments if needed.

It might be reasonable to implement a service action that would allow more flexibility (explicitly specify wait or not, and allow to interrupt wait with M108), but I am quite confused how to neatly register a service action taking into account config flow and multiple printers in a single integration.

NB: For X1E it is also possible to set chamber temperature, but it is done using a completely separate command, not with gcode, and I do not have a X1E unit to test that.

### DIFF
--- a/custom_components/bambu_lab/const.py
+++ b/custom_components/bambu_lab/const.py
@@ -16,6 +16,7 @@ PLATFORMS = (
     Platform.FAN,
     Platform.IMAGE,
     Platform.LIGHT,
+    Platform.NUMBER,
     Platform.SELECT,
     Platform.SENSOR,
     Platform.SWITCH
--- a/custom_components/bambu_lab/number.py
+++ b/custom_components/bambu_lab/number.py
@@ -1,0 +1,110 @@
+from dataclasses import dataclass
+from collections.abc import Callable
+
+from homeassistant.components.number import (
+    NumberEntity,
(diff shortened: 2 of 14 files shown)


AdrianGarside:
"but I am quite confused how to neatly register a service action taking into account config flow and multiple printers in a single integration." 

Each printer is handled by a dedicated instance of the integration so you never have to worry about multiple printers. There's no possibility of accidental cross-contamination.

AdrianGarside:
"P1*/A1* printers in LAN mode "set and wait" commands are used (because "instant set" commands are blocked)."

P1*/A1* printers in mixed mode (bambu cloud connection + local mqtt connection) are blocked. I would not add support for set and wait commands in that state - I would simply make the sensors unavailable for now to reduce risk of unintended interactions.

PR #696, opened 2024-11-06
Starting the camera is optional, defaults to on

kanflo:
The P1S can only handle two simultaneous image streams and this integration will use one of them. I have plans to use other instances of this superb library where the image stream is not needed, and would eat the last stream slot. This PR allows for specifying that the image stream should not be initialized, but of course defaults to `on` for compatibility.


AdrianGarside:
Awesome. I was planning to add this soon - the main pain will be making it a persistent setting (along with a persistent setting to use the older image pseudo-camera for folks that have browser problems with the new camera camera).

AdrianGarside:
Are you just making a local edit on your side to change the default?

kanflo:
I am currently using `pybambu` outside of HASS so that's where I can turn the camera stream off.

PR #726, opened 2024-11-10
Fixed ordering of code in get_tasklist, zapped whitespace

kanflo:
The check for `curl_available` was performed after the call to `curl_requests` in `get_tasklist`. Also zapped some white space in the vicinity,


AdrianGarside:
Thanks for catching this.

PR #762, opened 2024-11-18
Remove unsupported and increasingly broken wled blueprint

AdrianGarside:


Pharkie:
Will there be a new, working Blueprint? Or is it best to just set up our own automations manually?

AdrianGarside:
If someone wants to resurrect and fix it, I'm happy to add it back. But someone added it and then basically abandoned it and it's not correctly matched to the entity names (possible ever) or the fact that Bambu stopped firing a bunch of the states that it was trying to trigger off.

PaulBiod:
I forked it some time ago, and it works for me (P1S). If people are interested, I may improve it

PR #778, opened 2024-11-25
Adds a service to send arbitrary gcode commands to the printer

gpambrozio:
I want to add an automation to home all axis when I turn on my printer. 

This adds an Action to the integration where we can send arbitrary commands to it.

Not sure how this will behave if people have many printers. As I only have one it's not an issue for me.

Might also be good to add safeguards to not send the command when printing maybe???

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,6 @@ instead of OAuth.
 | Pause               |                                                  |
 | Resume              |                                                  |
 | Stop                |                                                  |
-| Home                | Homes all axis                                   |
 | Manual Refresh Mode | P1/A1 only and only available in local MQTT mode |
 
 ### AMS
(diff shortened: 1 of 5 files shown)


Skydev0h:
Wow, adding a service turned out to be so simple! Nice code to learn of.

> Might also be good to add safeguards to not send the command when printing maybe???

Actually, it is good this was not implemented, since sending some gcode while the printer is printing can be beneficial in some cases (change fan speed or nozzle / heatbed temperature, for example, I know that is implemented natively, but who knows, what gcode can be useful in the future).

northcoastam:
This sounds neat!

>sending some gcode while the printer is printing can be beneficial in some cases 

Does this mean I could send the Marlin M220 gcode to change the printing speed to rates other than the built-in "Silent/Sport/Ludicrous"?  I remember watching one tall, slender print (in TPU, I think it was) where near the top it was really starting to flex around, even after switching to silent mode.  I wished it was like my old Ender 3 Klipper printer where I could slow it down to any arbitrary speed.

PR #787, opened 2024-12-08
Allow different credentials

AdrianGarside:
Users have asked for the ability to use different credentials - either to allow multiple or because of an account change. The automatic re-use of existing working credentials blocked this. Add a new from in the config flow that allows users to opt not use the existing credentials. Choosing not to sends you to the login form with just the discovered user region pre-set.


Rupp3r:
This change is so awesome - thank you so much! 🙏🏻

Could it be a part of the Cloud/Lan options?
Maybe above or under the radio buttons?

The extra setup step is kinda meh 😅

AdrianGarside:
I'll think about that. Originally, I didn't because I didn't know if the user would be attempting to cloud login / have working credentials from another instance. But now that you've prompted me it occurs that I do know that - if they have an existing instance configured with cloud credentials and it's working then I can know that up front and without making unapproved calls to Bambu (since by definition the user was ok with that when they set up the other instance).

AdrianGarside:
@Rupp3r https://github.com/greghesp/ha-bambulab/releases/tag/v2.0.39-dev4

PR #791, opened 2024-12-14
fix: tray serial number

excalbian:
I noticed the attribute for the spool serial number wasn't consistent between tray 1 and the other 3 trays. While fixing `en` (english) I noticed the same issue appeared in other translations: `el`, `ko`, `pt`, `zh-Hans` (note that I don't speak any of these languages, just made the name for the `tag_uid` attribute consistent).


AdrianGarside:
Thanks for catching and fixing this.

PR #812, opened 2024-12-29
Update number.py

cinghialino:


custom_components/bambu_lab/number.py
```diff
@@ -43,8 +43,8 @@ class BambuLabNumberEntityDescription(NumberEntityDescription, BambuLabTemperatu
         set_value_fn=lambda device, value: device.temperature.set_target_temp(TempEnum.NOZZLE, value)
     ),
     BambuLabNumberEntityDescription(
-        key="target_bed_temperature",
-        translation_key="target_bed_temperature",
+        key="bed_target_temperature",
+        translation_key="bed_target_temperature",
```
t3chguy, line 47: Looks like this will break translations

custom_components/bambu_lab/number.py
```diff
@@ -43,8 +43,8 @@ class BambuLabNumberEntityDescription(NumberEntityDescription, BambuLabTemperatu
         set_value_fn=lambda device, value: device.temperature.set_target_temp(TempEnum.NOZZLE, value)
     ),
     BambuLabNumberEntityDescription(
-        key="target_bed_temperature",
-        translation_key="target_bed_temperature",
+        key="bed_target_temperature",
```
t3chguy, line 46: Looks like this makes it inconsistent with `target_nozzle_temperature`

cinghialino: sorry about that - I'm not a coder. the idea was to have more consistent naming 

`sensor.[printer]_bed_temperature` and `sensor.[printer]_bed_target_temperature` with `number.[printer]_bed_target_temperature`
or 
`sensor.[printer]_nozzle_temperature` and `sensor.[printer]_nozzle_target_temperature` with `number.[printer]_nozzle_target_temperature`

custom_components/bambu_lab/number.py
AdrianGarside: The problem is that these changes (much as I want to make them) break people's existing dashboards.

AdrianGarside:
Nice as this is, it breaks existing users' dashboards.

PR #840, opened 2025-01-20
Add Thai  lingo Translation

VaAndCob:
add thai translation


AdrianGarside:
@VaAndCob there is a syntax error in the json:
Error: R] [JSON] Invalid JSON file translations/th.json

PR #843, opened 2025-01-20
Drop cloud credentials when reconfiguring to lan mode

AdrianGarside:


var-au:
Doing god's work, thank you @AdrianGarside

PR #859, opened 2025-01-25
Add print_project_file action, print 3MF files

DanielsSt:
Needed a way to start print from HA, might be worth sharing.
In general, first tried with gcode_file , but could not get that working on my P1, but project_file seems to be working as expected.

Inspiration for MQTT payloads was taken from:
- https://github.com/Doridian/OpenBambuAPI/blob/main/mqtt.md#printproject_file
- https://github.com/mattcar15/bambu-connect/blob/main/bambu_connect/ExecuteClient.py#L41

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ See [device triggers](docs/DeviceTrigger.md).
 
 * **Send Command**: Sends arbitrary GCODE to the printer. Be careful as it does not check if the printer is running a job
 or not so before invoking this action, you should check the printer state to ensure it is not running a job.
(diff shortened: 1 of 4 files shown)


AdrianGarside:
@DanielsSt this hit a validation error:

Error: R] [SERVICES] Invalid services.yaml: min and max are required in slider mode for dictionary value @ data['print_project_file']['fields']['plate']['selector']. Got {'number': {'min': 1}}

DanielsSt:
@AdrianGarside Should be fixed now, at least `hassfest` check passed on my fork 👌

ndom91:
Is there any way to write remotely to the printers SD card? 

I'd love to be able to avoid having to pre-load the SD card with 3mf project files 🤔

AdrianGarside:
The esp32 that runs the P1/A1 printers is very low power. Max transfer rate is around 70kb/sec

PR #867, opened 2025-01-30
Fetch cover image, print length and weight via FTP in LAN only mode

### LINKED ISSUE #849: [Feature] Provide Cover/Preview Image through local LAN
### Describe the feature

Can the cover/preview image be pulled into Home Assistant via. local LAN, not cloud? I'm referring to the colored rendering of the product that's shown on the A1 screen during the print, not the camera entity.

Bambu's credential push leads me to believe that read-only values should be available over LAN. However this particular cover/preview image feature appears to be reliant on the cloud.

### What device is this for?

A1

### Other Information

_No response_

steveneaston:
## Changes
- Adds a "Use FTP" config switch
- Enables cover image, print length and weight sensors when FTP is enabled
- Fetches the 3mf via FTPS and sets sensors accordingly

## Details
I expect this may need some work, since I'm not too familiar with Python or building Home Assistant components in general. Areas I'm not clear on:

- Does FTP fetch need to be threaded so it doesn't block other processes?
- Used `tempfile` to download the 3mf into — _does HA have an approved method?_
- I assumed FTP is supported by all printers so hard-coded the feature as `True`
- Language support — _I've added `en`, but should I try and add the others?_
- Cover image content type is defined as `image/jpeg` but format is png (has transparency) — _doesn't seem to affect rendering_

Resolves #849.


WolfwithSword:
I wrote a standalone script about a year or so ago for python FTPS to the printer, so some notes I'll just add in case you run into some issues:

1) 

We copied from the same original stackoverflow post, though mine was edited to contain this method inside the new class, otherwise it didn't work reliably with all printer models.

```py
    def ntransfercmd(self, cmd, rest=None):
        conn, size = ftplib.FTP.ntransfercmd(self, cmd, rest)
        if self._prot_p:
            session = self.sock.session
            if isinstance(self.sock, ssl.SSLSocket):
                    session = self.sock.session
            conn = self.context.wrap_socket(conn,
                                            server_hostname=self.host,
                                            session=session)
        return conn, size
```

2) May beneficial to do an `nlist` (`LIST` does not work on all models) to see verify if it is available before downloading. X1 supports wildcards in the nlist, A1/P1 series do not.

3) A1/P1 series may have different file directory structure compared to the X1 series. Even more so if cache to sd card is enabled (and it preferably would be, otherwise *very* often, you cannot access a freshly sent print as it is in internal memory from the get-go unless cache print is enabled). So any FTPS feature should warn that sd card caching should be enabled for enhanced reliability.

As for paths, it can vary depending on print method, location, and printer series, and sometimes the path given in the request command isn't even accurate depending on print method. Essentially, it can be either at root `/`, or `/cache`, or `/mnt/sdcard` (which I think just goes to `/sdcard` when FTPSing). 

4) Also, if you "reprint" something via UI after it's been printed / print on last-print via printer UI, it will print off the internal memory version, which cannot be FTP'd. This is an edge case so a fallback option would be to either reuse last known values, or to keep in memory a story of 3mf file locations and their associated print request id's (which is what the print job in memory is named after). Took me far too long to figure this edge case out.

steveneaston:
Immensely helpful information @WolfwithSword, thank you. I'll try and get some of these changes implemented this weekend.

The changes this adds is specifically targeting LAN only mode, and from what I've seen with my P1S, BambuStudio and Orca upload the 3mf using FTPS to the SD card's root so the file is available immediately to retrieve. Otherwise, the integration pulls this data from cloud API.

My P1S was previously in cloud mode, and I noticed `/cache` had .3mf, .bbl and .gcode files for each print. I've been in LAN-only mode for the last couple of weeks, `/cache` has no new files and all prints are in `/` — .3mf only. I _assume_ SD caching is a cloud-specific feature, but I'd be keen to hear from X1 and A1 users what cards look like in LAN-only mode.

Couple of mentions of SD cache in the docs:
> This feature is enabled by default in Print Options. If you select this option when printing through cloud, the 3mf will be stored in the cache folder of the MicroSD Card root directory.
> [X1](https://wiki.bambulab.com/en/x1/manual/screen-operation)

> The first row is the SD card cache folder, which is used to cache historical print files. The others are internal model files.
> [P1](https://wiki.bambulab.com/en/p1/manual/screen-operation#folders-and-files)

Long story long, it sounds like a sensible idea to `nlst` root and a few other known directories in an effort to find the file.

Thanks again @WolfwithSword!

WolfwithSword:
>I _assume_ SD caching is a cloud-specific feature, but I'd be keen to hear from X1 and A1 users what cards look like in LAN-only mode.

SD Card caching is not cloud specific, but is a toggleable feature in the printer UI. It is used by cloud initiated prints for sure, but can also happen for LAN prints depending how it was sent from BBS/Orca.

Or at least, back when I did a ton of testing it was like that. It could have changed to *only* do it when initiated from cloud maybe. Unsure but probably not a bad idea to check anyways in the event the nlist from root directory doesn't find it.


> Long story long, it sounds like a sensible idea to `nlst` root and a few other known directories in an effort to find the file.

Yeah, this is likely the safest approach \o/ nlist root, then cache, and then sdcard (it depends on printer model but as long as it's in a try catch, can just do all 3). Still sucks that wildcard filtering like for *.3mf only works for X1, so filtering filetype can come afterward.

AdrianGarside:
For the new string, you can automate localization to all other supported languages by doing:
cd scripts
python3 auto_translate.py

AdrianGarside:
Generally looks great. Needs a little work to make the ftp interaction asychronous so it doesn't block the mqtt thread. I can work on that. And then we'll need to give it some love to better handle all the possible paths for the model file.

steveneaston:
Thanks @AdrianGarside. I've made the changes now, with it looking in `/`, `/cache`, `/models` and `/sdcard`. `/models` is a path on my P1S card which came preloaded with a few models from Bambu.

AdrianGarside:
So I was looking to see what it would take to show the last print details when a print is no longer in progress (and gcode_filename is empty. Looking at my machines, the latest timestamp in the cache directory was my last print.

And I also the image was already extracted (even on the P1* printers) to the image directory:
image

This might be a more efficient thing to grab first so that we aren't blocked on downloading the 3mf file - which can be very large for complex prints.

AdrianGarside:
I have an sd card connected to my X1C but there is no /sdcard directory

WolfwithSword:
> I have an sd card connected to my X1C but there is no /sdcard directory

the /sdcard directory is a P1/A1 specific thing. Not shared/consistent with X1

steveneaston:
`/sdcard` isn't present on my P1S when the card is mounted.

> And I also found the image was already extracted on the P1P/S printers (but not X1C) to the image directory.

I saw these but they are only 128x128px (on mine). Could maybe work as a quick preview before being replaced by the larger images async?

AdrianGarside:
I have no /sdcard directory on my P1P/S - the entirety of the files are on the sd card so I'm not sure what /sdcard would be?

AdrianGarside:
> I saw these but they are only 128x128px (on mine). Could maybe work as a quick preview before being replaced by the larger images async?

Oh yes, I hadn't noticed they were such low res. Not really worth using then unless we start to get complaints it takes to long for the real cover image to appear after print start.

AdrianGarside:
@WolfwithSword 
"May beneficial to do an nlist (LIST does not work on all models) to see verify if it is available before downloading"

Which models don't support LIST? We need that to get the timestamp so that we can load the data for the last printed model when the printer is idle.

d-nx:
Would it be possible to also grab your timelapse vid with this method?

AdrianGarside:
> the /sdcard directory is a P1/A1 specific thing. Not shared/consistent with X1
None of my P1 have it. I'm not sure what it would be given the sdcard is the entirety of the files system on those.

AdrianGarside:
> Would it be possible to also grab your timelapse vid with this method?

Yes. That hard part is making a usable experience for doing that in HA.

d-nx:
> > Would it be possible to also grab your timelapse vid with this method?
> 
> Yes. That hard part is making a usable experience for doing that in HA. What do you want to do exactly?

If the integration is connecting to the printer via FTP, I would hope to either be able to browse the printer/timelapse directory and download a specific timelapse.avi, or maybe the integration helps you start downloading the last timelapse.avi file 

Hope this makes sense, Ive had success using the following command (macos terminal):
`curl --ssl-reqd --ftp-ssl-control -k -u username:password "ftps://printerIP:990/timelapse/" | awk '{print $9}' | sort -r | head -n 1 | xargs -I {} curl --ssl-reqd --ftp-ssl-control -k -u username:password -o ~/Downloads/{}_$(date +%Y%m%d%H%M%S) "ftps://printerIP:990/timelapse/{}"`

PR #876, opened 2025-02-02
Enable TLS certificate validation

### LINKED ISSUE #872: [Feature] SSL cert usage take #2
### Describe the feature

See suggestions in https://github.com/greghesp/ha-bambulab/pull/809

### What device is this for?

A1

### Other Information

_No response_

j4k0xb:
Closes #872 

This PR validates the TLS certificates and host names¹ for:
- [x] MQTTS cloud
- [x] MQTTS local
- [x] FTPS local
- [x] camera stream local

which should completely prevent spoofing/man-in-the-middle attacks.
¹: either the cloud host name (like `us.mqtt.bambulab.com`) or the printer's serial number (like `01P***********5`)

I only tested the individual changes directly with a small python script for local/cloud so far.
Would appreciate if someone can test if it also works within HomeAssistant.


AdrianGarside:
It works for me even without the wrapper class.

j4k0xb:
Wrapper is needed because `check_hostname` is no longer disabled

AdrianGarside:
> Wrapper is needed because `check_hostname` is no longer disabled

Ah, I missed some of your changes to the tls setup. Now I get an error:

Exception. Type: <class 'ssl.SSLCertVerificationError'> Args: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: IP address mismatch, certificate is not valid for '10.10.10.81'. (_ssl.c:1000)

j4k0xb:
Ok adding it back in, ig `._serial` isn't always set so checking wouldn't make sense

AdrianGarside:
_serial should always be set (can't communication with a P1/A1 printer without it) and I checked it had the correct values.

AdrianGarside:
@j4k0xb This is still a draft PR - are more fixes still needed or should I try this out again?

j4k0xb:
I made more changes yesterday, should be done now so please try it again

AdrianGarside:
Unfortunately it's still failing as the servername trick still doesn't seem to be working:
```
2025-02-04 09:53:04.460 ERROR (P1P-Mqtt-147) [custom_components.bambu_lab.pybambu] Exception. Type: <class 'ssl.SSLCertVerificationError'> Args: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: IP address mismatch, certificate is not valid for '10.10.10.81'. (_ssl.c:1000)
```

j4k0xb:
Weird, works on my machine...
Then keep the certificate but add back `ctx.check_hostname = False` for LAN only? It's not fully secure but at least users won't get these SSL errors.

AdrianGarside:
I think that’s fine. There’s no meaningful security benefit since the end user chose to talk to the specific IP address when setting up the integration. So to exploit this a hacker has to be able infiltrate your network take over that IP address and run a compatible but somehow malicious Mqtt server on it that can attack the caller. Or do the same with a man in the middle attack.

But if they are able to do all of that then it was long ago game over - they already own your network.

j4k0xb:
>  So to exploit this a hacker has to be able infiltrate your network take over that IP address and run a compatible but somehow malicious Mqtt server

In practice it's more like: they just have to be on the same network, spoof all IPs and run a generic server that logs everything (including your printer's access code).

The initial thought behind the PR was that it protects even against such attacks. You'd maybe have worry about your other devices but the printer couldn't be compromised.

AdrianGarside:
Once they're on your network, you're already screwed. The printer is likely the least of your problems.

j4k0xb:
Yeah but it's about minimizing damage. ha-bambulab can improve the situation for printers and if other software is also secured then you might not even be screwed anymore in the future.

j4k0xb:
> if you're defending the window when the attacker just came in through an open door

*an attacker who first has to buy a 3D printer and manage to extract the keys from it. Already greatly reduces the attack surface but agreed, effectively it won't prevent someone determined.

spali:
Hi @AdrianGarside & @j4k0xb 
Just my 2 cents about this in case the local variant get's another shot.
First love it when something gets more security. But in the local case, please keep in mind that lan-mode only will probably fail in the future if certificate is validated. Due the recent "ugly BBL topic" (which is in my eyes one of the main driver for current new users of this integration, including me), a lot of people go stealth with their printer) and BBL mentioned certificates may expire without any connection to the internet by BambuStudio or the printer itself.
Thats would be the case with my current workflow (using orca slicer and the printer is only allowed (internet) to get it's time by NTP). 
**If you make another attempt** I would highly appreciate if it would be at least possible to disable validation for the local case if the certificate expires in the future.

j4k0xb:
I understand your concern about it breaking some time in the future, here's some clarification:

> and BBL mentioned certificates may expire without any connection to the internet by BambuStudio or the printer itself.

- Certificates in the printer: connection to the cloud. However, on many embedded system this expiry check is disabled by default. Would be interesting if someone can confirm this by setting the printer's time a few years ahead.
- Certificate in Bambu Studio, OrcaSlicer, ha-bambulab etc.: connection to the printer.

> Thats would be the case with my current workflow (using orca slicer and the printer is only allowed (internet) to get it's time by NTP).

The expiry check is only performed on your your PC, regardless of having internet access or not.
OrcaSlicer would also run into this problem: https://github.com/SoftFever/OrcaSlicer/blob/main/resources/cert/printer.cer

> If you make another attempt I would highly appreciate if it would be at least possible to disable validation for the local case if the certificate expires in the future.

Definitely. Future = `01/04/2032 03:42:11` to be exact, so there's plenty of time to either update it or add a setting.

PR #897, opened 2025-02-03
Skip object support

### LINKED ISSUE #866: [Feature] Support "skip object" functionality
### Describe the feature

**WHY:**
The P series currently must use the app/be cloud-connected in order to use the skip objects feature. This is obviously a huge issue going forward with Bambu's firmware stance and wanting to go offline. Implementing support for skipping objects locally would be greatly appreciated.
**HOW:**
It seems someone has [done the sniffing](https://github.com/Doridian/OpenBambuAPI/pull/31/files) to find the MQTT commands to skip objects. This should be relatively simple to re-implement object skipping commands. The complexity will come from creating a supporting frontend card to allow the user to select the objects to skip. Generating that thumbnail/polygons will be a relatively complex task as it will require access to the print file/GCode (presumably via the [inbuilt FTP](https://github.com/Doridian/OpenBambuAPI/blob/main/ftp.md)), then parsing the files oultlined in [this post](https://www.reddit.com/r/OpenBambu/comments/1ibt75f/comment/m9x5kzr/).

### What device is this for?

P1S

### Other Information

Unfortunately this is above my head, but am happy to contribute to a bounty for such a feature to be implemented.

steveneaston:
## Changes
- Adds service call/action "Skip objects"
- Adds image sensor "Pick image"
- Adds generic sensor "Printable objects"
- Adds config switch "Label pick image"

## Details
Retrieves the object "pick" image from the printer FTP (when enabled) and stores each object's identify ID and name in the "Printable objects" sensor's attributes. If "Label pick image" is enabled, the pick image is labelled with the ID against each object.

| Pick image | Printable objects attributes |
| --- | --- |
| Object picklist labelled with identify IDs | Printable object sensor attributes |

Send skip object command using the action:
```yaml
action: bambu_lab.skip_objects
data:
  objects: "416"
```

## Caveats
- I had an iteration where the bounding box for each object was drawn round it, however I discovered the bounding box coordinates from `plate_{x}.json` had an offset when the plate wasn't 1. Some offsets affected x1 and x2, some y1 and y2, some affected both.

- This addition requires the `pillow` lib but I didn't add it to the manifest requirements because it seems to run without it.

- Ideally, there would be a UI so the user could click an object to cancel it. That's out of scope for this PR, but could be integrated using the new Pick Image sensor along with @thunderbolt93's [suggestion](https://github.com/greghesp/ha-bambulab/issues/866#issuecomment-2629344321).

- I originally had this compiling the printable objects into a Todo List sensor. The user could click the todos as "completed" and that would cancel them. @AdrianGarside I have those changes stashed, let me know if that'd be a better solution or addition to what's provided here.

- Bambu Lab have restrictions on [skip object](https://wiki.bambulab.com/en/general/skipping-objects#matters-needing-attention), limiting commands to >1 <=64 objects. I haven't restricted the service command to account for that.

	> 3. If there is only 1 object on the plate, or if there are more than 64 objects, the Skip object function cannot be enabled.

Resolves #866


AdrianGarside:
"I originally had this compiling the printable objects into a Todo List sensor. The user could click the todos as "completed" and that would cancel them."

We're looking into how to combine the existing ha-bambulab-cards repo into this one so that this integration installs its own cards. That's all looking promising. I'd then want to try and build a card for this that has the more complex visual UI and hopefully can trigger the service call itself. But all of that's new to me so even assuming we don't hit blockers I expect it'll take a while to build that.

What did the todo list experience look like?

drkpxl:
Just an idea could we use text labels? So in Orca Slicer each object is named and you check the names you want against a checklist and hit an "Exclude" cta? May be an easier V1 and I can add it to the [printwatch](https://github.com/drkpxl/printwatch-card) card pretty easily?

ichisaur:
No, object names aren't unique and that implementation would still require some confirmation step for checking which object is selected. 

For what it's worth, the current BBL Handy app implementation has I believe both a touch UI for selecting the object directly, as well as a check list with objects names. The check list highlights the objects as red before you click the cancel button to confirm. I think that's very similar to the approach I described. The only issue with HA is making sure users are aware of needing to hit a confirm button. 

Maybe calling the preview,select list, and cancel button: Selection Preview, Select Object, and Cancel Selected Objects could help users work out the process?

drkpxl:
Screenshot 2025-02-04 at 3 44 28 PM
Did this real quick as example?

ichisaur:
> Screenshot 2025-02-04 at 3 44 28 PM
> 
> Did this real quick as example?

Yeah the problem is you can have 64 objects named cube and not know which one you need to cancel. 

Cube_1, cube_2 etc is also not very helpful without a visual reference since the default orca plater does not organize parts in any logical order.

steveneaston:
The idea behind this PR was purely to provide the data. I think for something interactive like this, where the action of cancelling an object involves several entities having to work together (reference image, list of selected objects, button) we really need a separate card feeding off the data the integration provides.

> Just an idea could we use text labels?

The IDs and object names are in the Printable Objects sensor's attributes, so in the printwatch card, loop through each of those. If that's not the best storage they could be added in a single attribute as a comma-separated string — or maybe two attributes, one with identify_ids, one with object_name?

AdrianGarside:
I think the only somewhat good experience we can have for this will require a card where we have the freedom to craft a similar UX to that in the Bambu software. I am confident the visuals could be built (although if I'm doing it, it'll take me a while to leave javascript/css :)) but I don't know yet if the card can then trigger service calls. I assume and believe it can but I'd also have to work out how to do that at the end once the user's choices have been made.

ichisaur:
Skipped objects are provided in the pushall MQTT statement as s_obj list.

AdrianGarside:
Thanks. I thought I recalled seeing something like that.

PR #873, opened 2025-02-02
Load latest model data when not printing

AdrianGarside:
@steveneaston made some changes on top of your addition

Issue #898, opened 2025-02-03
[Bug] FTP model download is not updating correctly on print start (P1/A1 only?)


AdrianGarside:
### Home Assistant Version

-

### Bambu Lab Integration Version

-

### Describe the bug

See #894 - it updates correctly on integration start but not properly on print start. Maybe the model isn't fully uploaded to the printer at that time with a P1/A1 where it might be with an X1.

### To Reproduce

.

### Expected Behaviour

.

### What device are you using?

A1

### Diagnostic Output

```shell
,
```

### Log Extracts

```shell

```

AdrianGarside:
@steveneaston FYI

steveneaston:
I'll have a dig into this tomorrow morning and trigger off a few prints, see how each part of the `_update_task_data` process reacts.

FWIW, if the file isn't available until a little while after the print event is fired, [this](https://github.com/greghesp/ha-bambulab/pull/905/commits/1bed1a0cc92a655c8e8d8f91b0e0e869404d742a) will definitely help all the other sensors from crashing out. It was throwing an exception while I was testing something else.

gcs8:
moved to beta, it picked up the cover image just fine at 70%

![Image](https://github.com/user-attachments/assets/e5e39a24-a6d7-46c6-816d-f8731b926702)

steveneaston:
I've done some testing with OrcaSlicer and the latest changes on `main`. No problems fetching the 3mf or extracting it, printer isn't blocked during the process. I'm going to do some tests with the same three models using BambuStudio this afternoon.

---

### Print 1/3
**3mf**: 40kb
**gcode**: 200kb
**Fetch**: 2 seconds

```
2025-02-04 12:24:17.617 DEBUG (P1S-Mqtt-125) [custom_components.bambu_lab.pybambu] GCODE_STATE: FINISH -> PREPARE
2025-02-04 12:24:17.617 DEBUG (P1S-Mqtt-125) [custom_components.bambu_lab.pybambu] GENERATED START TIME: 2025-02-04 12:24:00
2025-02-04 12:24:17.618 INFO (P1S-FTP-131) [custom_components.bambu_lab.pybambu] Updating task data by FTP
2025-02-04 12:24:17.621 DEBUG (MainThread) [custom_components.bambu_lab] BUS EVENT: event_print_started: {'device_id': '[REDACTED]', 'name': '', 'type': 'event_print_started'}
[...]

PR #905, opened 2025-02-04
Atttempt to load previous model from subtask_name if available

steveneaston:
## Changes
- Fixes issue where FTP attempts download when no file was found
- Attempt to find original .3mf from `subtask_name` value if present. This value often persists in the pushall MQTT response unless the printer is restarted.
- Adds FTP directory caching

## Details
I noticed after #873 every time Home Assistant was restarted it was loading an old model rather than the last one I printed.

My P1S `/cache` directory looks something like:
```
Jan 21 12:34 AAAA_plate_2.gcode
Jan 20 10:20 BBBB_plate_1.gcode
Jan 20 17:02 CCCC_plate_1.gcode
Jan 19 11:25 DDDD_plate_6.gcode
Jan 19 11:25 DDDD.bbl
Jan 19 11:25 DDDD.3mf
Jan 18 21:27 EEEE_plate_3.gcode
Jan 18 21:27 EEEE.bbl
Jan 18 21:27 EEEE.3mf
```

Jan 19th was the last time I did a cloud-based print through OrcaSlicer. Since then, it's been LAN only. Both OrcaSlicer and BambuStudio upload prints to the FTP root, so aren't picked up by the "find latest by date" search.

---

The changes here updates the "find latest" method to defer to the `subtask_name` value when `gcode_file` is empty after Home Assistant restarts. The subtask name is often included in the pushall MQTT, however it is cleared if the printer is power cycled _(there may be other events that clear it)_. The subtask name is the name of the file without any extensions, e.g. `{"gcode_file": "Foo bar.gcode.3mf", "subtask_name": "Foo bar"}`.

The FTP caching retains a list of path contents (from `/` and `/cache`) until it can't find the model it's looking for. After which it'll clear each directory's cache sequentially until it finds it. So if nothing new is ever added to `/cache` but files are re-printed from it, it shouldn't have to fetch the contents each time.


AdrianGarside:
My X1 doesn't expand the 3mf at all: 
image

AdrianGarside:
Not all print files have .stl in them on my P1 printers:
image

steveneaston:
Thanks for sharing those @AdrianGarside. The plot thickens. I think, in that case, I'll discard the .gcode → .3mf search, and see if I can improve the 3mf lookup.

While I was running tests for #898, I saw that both OrcaSlicer **and** BambuStudio upload the files via FTP to `/`, not `/cache`. Also, BambuStudio isn't creating a `.bbl` file, so I assume that comes only via cloud. Were the prints you pushed [here](https://github.com/greghesp/ha-bambulab/pull/905#issuecomment-2632482340) via cloud or FTP?

> Older prints still have no 3mf at all - just the gcode. Maybe they got cleaned up?

That sounds right, from the [X1 docs](https://wiki.bambulab.com/en/x1/manual/screen-operation#print-cache) they mention old 3mf files being deleted:

> The print cache can store up to “8” 3mf files. If it exceeds, the oldest 3mf file cached will be deleted. When the remaining capacity of the MicroSD Card is less than 500MB, some old print cache 3mf files may be deleted to ensure that the current print 3mf file can be cached.

Can't find any mention of it on the P1/A1 docs, but I would expect those do some file management as well.

AdrianGarside:
I have 21 3mf files in /cache on my P1P and one of my P1S. And 20 3mf files on the other P1S. I only have 8 in /cache on my X1.

None of my examples were pushed by ftp - I'm still running in cloud mode now except for one printer I have in lan mode to test I haven't broken anything but I haven't sent anything to it yet.

spali:
I was curious, so tried the beta for my X1C.
Not sure if it helps or is already known... I started a print directly from the sdcard (not cache) throught the X1C itself and it shows a image of another print from the past.

AdrianGarside:
> I was curious, so tried the beta for my X1C. Not sure if it helps or is already known... I started a print directly from the sdcard (not cache) throught the X1C itself and it shows a image of another print from the past.

That's probably this issue that we're trying to fix. Or at least more of the same issue. There might be a gap that if the file only exists as gcode we don't have access to the 3mf to get the data we want so we may need to find a way to detect that and make the sensors unavailable if we can't put correct data into them when in that situation.

spali:
For my original Bambu+Lab+Reusable+Spool+Hex+Snowflake.3mf I printed by sending (not printing) it to the sdcard. I reuse it then always directly from the printer if I need to print a new spool.
Connected via SFTP, I have a file `Bambu+Lab+Reusable+Spool+Hex+Snowflake.gcode.3mf` in the root. I assume this is the gcode. So I can confirm.. there is no non gcode 3mf file on the sdcard available... but I would also assume somewhere the printer has more information saved... as the picture is displayed on the printer and I also can choose on the printer which plate I want to print... so some metadata must be there too.

AdrianGarside:
That case could be tricky to get right since I suspect in the existing file reuse case (especially on the X1 where it didn’t unzip the 3mf file contents) there is no timestamp change to determine (after the print completes) that it occurred. We’ll have to compare timestamps on cached locally downloaded files on HA to determine if they are better than the latest file timestamp on the printer.

steveneaston:
@AdrianGarside I wonder if we can't see the forest for the trees here. I just booted back up my dev environment, and noticed that even though `gcode_file` is empty, `subtask_name` persists and it's the name of the last printed 3mf without the extension (extension**s**, sometimes the extension is `.gcode.3mf`, sometimes just `.3mf`, subtask_name omits both).

If we took that, and matched it against a simpler nlst search of `/` and then `/cache` if necessary, it's going to be far more efficient than using LIST and parsing dates. And if it can't find it… not the end of the world to not be able to see the last print?

AdrianGarside:
That does sound a lot simpler certainly as an optimization. Looking at my idle machines, I see:

P1P in Lan mode, no prints since I switched it over:
      "subtask_name": "",
P1S:
      "subtask_name": "Measuring Bottle Lid.stl",
P1S:
      "subtask_name": "Walking Cat.stl",
X1C:
      "subtask_name": "",

Where gcode_file is always empty (since it's no longer printing):
      "gcode_file": "",

So in some cases it gives us the most recent print allowing us to directly locate the 3mf file. But in other cases we still don't know.

steveneaston:
This is working for me. Finds the latest print I made. I expect it'd work in cases where the print was actioned on the printer, too? Assuming `subtask_name` gets sent back in the mqtt payload.

```py
def _find_latest_file(self, ftp, path, extensions: list):
    def path_matches_subtask(filename):
        return filename.endswith(".3mf") and filename.startswith(self.subtask_name)

    for search_path in ['', '/cache']:
        try:
            path_contents = ftp.nlst(f"{search_path}")
            matches = list(filter(path_matches_subtask, path_contents))
            if len(matches):
                return matches[0]
        except:
            pass
    
    return None
```

Can look to combine the directory search in `_find_model_path` and look into caching, so it's only ever searching `/` and `/cache` once.

steveneaston:
@AdrianGarside I've made some changes and added FTP directory cache but I need to do a bit more testing tomorrow. I'll let you know when it's ready to review.

spali:
May be already known... but just found out... for the *.gcode.3mf files which I have almost exclusively since I print local. You can unzip them and get metadata including pictures of the plate in different orientation or formats. 
Maybe this would allow to have model image also for sdcard based prints?

steveneaston:
@AdrianGarside ready for you to have another look. I've updated the PR notes to explain a little about the caching. I've tested this with printing a file from Orca, then re-printing a second model directly from the device. It doesn't re-fetch the directory contents the second time.

I've also replaced some of the changes just made in #932, it's trying to search for `{self.subtask_name}.3mf` but with files printed locally, they're uploaded by Orca and BambuStudio as `Foo Bar.gcode.3mf` with the subtask name `Foo bar`.

AdrianGarside:
FYI - I pushed a quick fix this morning that will probably cause conflicts with your changes.

steveneaston:
Was that #932? I resolved them just before I commented. However this PR's now got merge commits from fork origin sync. If you need me to force push a rebased version I'll do that. But I'll wait to see what you think in case there's more changes to make

AdrianGarside:
> Was that #932? I resolved them just before I commented. However this PR's now got merge commits from fork origin sync. If you need me to force push a rebased version I'll do that. But I'll wait to see what you think in case there's more changes to make

Yes. No need to push a rebased version - I always squash merge contribution PRs otherwise the git graph becomes horrific very quickly. I'm at work so can't really look much for 6 or so hours.

PR #927, opened 2025-02-05
Gracefully handle download task error

greghesp:
Please test this one properly before merging :)

I think I've done it correctly, but I still can't get the functionality fully working on my side

### DIFF
--- a/custom_components/bambu_lab/pybambu/models.py
+++ b/custom_components/bambu_lab/pybambu/models.py
@@ -3,10 +3,12 @@ import math
 import os
 import re
 import threading
+import ftplib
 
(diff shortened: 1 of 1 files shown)


AdrianGarside:
The change is good as-is. The only question is whether it's sufficient (you're seeing a bare gcode file in the taskname which is why it didn't work for you - it requires the 3mf). The implementation of this is a bit of an onion peeling exercise to find out all the ways orca/bambu studio/reprints/makerworld prints/lan mode affects the file structure on the P1/A1/X1 printers. There's a lot of possible combinations to try to handle.

PR #933, opened 2025-02-06
Remove AMS filament mapping when using device print data

steveneaston:
## Changes
- Removes AMS print weight and length mapping

## Details
The filament data provided in the slice info JSON is not sufficient to map weight or length against a specific AMS tray or external spool. It does still let us tally _total_ print length.

Example filament node:

The following was printed from AMS tray 4, which was a SUNLU PLA+ in black. I deliberately set the colour in the slicer `#F500C7` to double check it wasn't switched after AMS mapping during print.
```
<filament id="1" tray_info_idx="P60ec010" type="PLA" color="#F500C7" used_m="2.27" used_g="6.78" />
```

I'm still scouring the rest of the 3mf metadata to see if the slicer's storing mappings in there but no luck yet.

### DIFF
--- a/custom_components/bambu_lab/pybambu/models.py
+++ b/custom_components/bambu_lab/pybambu/models.py
@@ -873,16 +873,12 @@ class PrintJob:
                             if metadata.get('skipped') == f"false":
                                 printable_objects[metadata.get('identify_id')] = metadata.get('name')
                         elif (metadata.tag == 'filament'):
-                            # Filament used for the current print job. The plate info does not distinguish
-                            # between AMS and External Spool, both AMS Tray 1 and External Spool have
-                            # an ID of 1
-                            LOGGER.debug(f"AMS Tray {metadata.get('id')}: {metadata.get('used_m')}m | {metadata.get('used_g')}g")
-                            
-                            # Print weights and lengths expect zero-indexed allocation, reduce the ID by 1
-                            ams_index = int(metadata.get('id')) - 1
-                            self._ams_print_weights[ams_index] = metadata.get('used_g')
-                            self._ams_print_lengths[ams_index] = metadata.get('used_m')
-                            
+                            # Filament used for the current print job. The plate info does not contain
+                            # enough data to map it to an AMS tray or external spool. Only the brand
+                            # ID and type are included, the color is whatever was set in the slicer
+                            # and not what is reported by the AMS.
+                            LOGGER.debug(f"Filament 1: {metadata.get('used_m')}m | {metadata.get('used_g')}g")
+
                             # Increase the total print length
                             print_length += float(metadata.get('used_m'))
                     


AdrianGarside:
That's odd. I am seeing the correct id. My latest test print was printing from tray 4:
    <filament id="4" tray_info_idx="P9816594" type="PLA" color="#161616" used_m="1.18" used_g="3.52" />
Are you fully synchronized when you printed and didn't have to remap at all when sending the print?

steveneaston:
> Are you fully synchronized when you printed and didn't have to remap at all when sending the print?

Yes, but that's part of my point, we don't know how people will assign filaments. Here's a better example:

I've got 4 filaments in my AMS, tray 2 is grey `#9B9EA0`, tray 3 is black `#000000`
Screenshot 2025-02-07 at 13 53 07

In my object assignment, I assigned cube 1 to tray 2 (grey), cube 2 to tray 3 (black). I assigned cube 3 to tray 2 and reassigned the colour as red `#FF0000`, and cybe 4 as tray 3 reassigned to green `#00FF00`
Screenshot 2025-02-07 at 13 55 51

When I go to print, the mapping guessed red as tray 1, green as tray 4 so I remapped them to tray 2 and 3 respectively.
Screenshot 2025-02-07 at 13 57 02

The print lists filaments as so:
```
<filament id="1" tray_info_idx="GFA01" type="PLA" color="#9B9EA0" used_m="2.22" used_g="7.06" />
<filament id="2" tray_info_idx="GFA01" type="PLA" color="#000000" used_m="2.23" used_g="7.09" />
<filament id="3" tray_info_idx="GFA01" type="PLA" color="#FF0000" used_m="2.32" used_g="7.36" />
<filament id="4" tray_info_idx="GFA01" type="PLA" color="#00FF00" used_m="2.33" used_g="7.39" />
```

You can see that the filament IDs are just in the order they appear under "Filaments" in the slicer, not necessarily how they were mapped to the AMS. And we couldn't accurately match up the tray idx with what's in the AMS because if the model sets a different colour, it may not be what colour the AMS has.

A lot of models I download have a handful of filaments added each with a unique colour to help identify parts. When I switch the filament I don't typically re-arrange the filament list to exactly match my AMS.

AdrianGarside:
Darn. Maybe the gcode has enough information for us to work it out. But for now removing them is the right call.

steveneaston:
Ah, maybe some good news. AMS mapping is done via the MQTT print command. So, I just printed with:
| Filament (slicer) | AMS tray |
| -- | -- |

steveneaston:
@AdrianGarside Ah, maybe some good news. AMS mapping is done via the MQTT print command. So, I just printed with:

| Filament (slicer) | AMS tray |
| -- | -- |
| 1 | 3 |
| 2 | 2 |
| 3 | 4 |
| 4 | 1 |

In the "print" report response, we're getting:

```id
"ams_mapping": [
  2,
  1,
  3,
  0
]
```

So, with zero-indexing, ams_mapping[0] (Filament 1) is AMS[2] (Tray 3) etc.

Doesn't look like we're loading these into the print job yet, but I'll have a look at getting it added back in now I know how they're mapped.

PR #940, opened 2025-02-08
Alpha version of Skip Object card.

AdrianGarside:
Functional but basic at this point.


AdrianGarside:
PR has moved to the separate cards repo as we decided to separate them and just coordinate injecting new versions of the card into this repo for releases with this integration but also allow for independent release from the other repo.
https://github.com/greghesp/ha-bambulab-cards/pull/1

PR #981, opened 2025-02-11
Extrude and retract filament service actions

steveneaston:
## Changes
- Adds service command (actions) to extrude or retract filament

## Details
Action contains a soft block to prevent extrusion or retraction if the nozzle temperature is below 170ºC. This is the same threshold applied by [Bambu and Orca](https://github.com/SoftFever/OrcaSlicer/blob/c21b044a9c566db272adaee7d88bfd7198c2a572/src/slic3r/GUI/StatusPanel.cpp#L30). As those applications also have, the action has a "force" boolean that will let the user override it.

The distance and speed are hard-coded to 10mm and 900 respectively. If it would be advantegous to accept a dynamic distance, I'll update it per #971 (with hard limits).

### DIFF
--- a/custom_components/bambu_lab/coordinator.py
+++ b/custom_components/bambu_lab/coordinator.py
@@ -191,6 +191,7 @@ class BambuDataUpdateCoordinator(DataUpdateCoordinator):
         force = data.get('force')
 
         if move not in ['EXTRUDE', 'RETRACT']:
+            LOGGER.debug(f"Invalid extrusion move '{move}'")
             return False
         
         nozzle_temp = self.get_model().temperature.nozzle_temp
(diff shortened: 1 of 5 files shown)


custom_components/bambu_lab/coordinator.py
```diff
@@ -177,6 +180,34 @@ def _service_call_send_gcode(self, event: Event):
...
+        move = data.get('type').upper()
+        force = data.get('force')
+
+        if move not in ['EXTRUDE', 'RETRACT']:
+            LOGGER.debug(f"Invalid extrusion move '{move}'")
+            return False
+        
+        nozzle_temp = self.get_model().temperature.nozzle_temp
+        if force is not True and nozzle_temp < 170:
+            LOGGER.debug(f"Nozzle temperature too low to perform extrusion: {nozzle_temp}ºC")
```
AdrianGarside, line 199: Make this an error or maybe an info would be enough so there's a persistant explanation in the logs for why the command got dropped. We don't want incorrect usage to just be silently ignored.

AdrianGarside (changes requested):
Minor fix needed.

AdrianGarside (approved):

Issue #533, opened 2024-05-30
[Feature] Extrude/retract (or load/unload) filament commands


WayneManion:
### Describe the feature

I have my printers on shelves. One of my printers is a P1P with the P1S conversion kit. Sometimes I have to turn the printer around to be able to load filament. When the printer is turned around I no longer have easy access to the controls on the front of the printer. 

On my Klipper printers I have macros set up to heat to 240C, turn off the nozzle heater, retract 25mm, extrude 50mm, and extrude 200mm. I have a Zigbee scene remote connected to Home Assistant that can trigger these macros. This means I can heat the nozzle and then retract or extrude filament as needed to perform filament (un)loading without access to the printer's display.

I would like to be able to do something like this with my P1P/S. I can already heat or cool the nozzle with a Zigbee remote, but I don't know if there is any way to provoke an extrusion or retraction move by using MQTT or the Bambu integration. 

Yes, I could use my phone, but when I have the printer turned around and one hand trying to push filament into the reverse Bowden tube I would rather just press a simple physical button. 

### What device is this for?

P1S

WayneManion:
I did some experimentation. I could not figure out how to get the TLS working with the mosquitto_pub client, so I used Node-RED. 

It is completely possible to send lines of gcode to the printer. At least when the printer is in LAN mode. Send a message to device/{{SERIAL}}/request like this:
```{
   "print":{
       "sequence_id":"0",
       "command":"gcode_line",
       "param":"G1 E50 F600 \n \n",
       "user_id":"1234"
   }
}```

And the machine will extrude a little bit of filament if the nozzle is up to temperature.

leah-potato:
I would LOVE a button where I can tap load / unload in home assistant. I don't have an AMS and I'm constantly switching stuff out - Wayne and I can't be the only ones keen for this?

PR #988, opened 2025-02-12
Load and unload filament service actions

steveneaston:
## Changes
- Adds service command (actions) to load and unload filament (#533)
- Adds `compare_version` utility method to compare two software versions
- Adds Device method `supports_sw_version` to check its software is equal to or newer than the given version
- Adds `AMS_SWITCH_COMMAND` feature to check Device's software supports `ams_change_filament` command

## Details
This is an initial version to provide filament unloading and loading from AMS and external spool. It uses the MQTT `ams_change_filament` command which is available after the firmware versions listed below. It's possible to load and unload on devices with older software, but I can add that in a follow-up PR.

The `compare_version` utility performs a simple match against two given versions to determine which is greater. `supports_sw_version` uses it to ensure the device is of at least _this_ version.

After the new filament is loaded, the nozzle temperature is set to the temperature provided, otherwise it attempts to find the midpoint between the min and max temperatures defined in the filament profile.

_MQTT command available since:_
| Device | Firmware |
| -- | -- |
| A1 | Always |
| A1 Mini | Always |
| X1E | Always |
| X1 | `01.05.06.01` |
| X1C | `01.05.06.01` |
| P1S | `01.02.99.10` |
| P1P | `01.02.99.10` |

_Version flags available in OrcaSlicer's [device profiles](https://github.com/SoftFever/OrcaSlicer/tree/main/resources/printers)._

Screenshot 2025-02-12 at 20 52 14

### DIFF
--- a/custom_components/bambu_lab/coordinator.py
+++ b/custom_components/bambu_lab/coordinator.py
@@ -29,12 +29,15 @@ from .pybambu.const import (
     SEND_GCODE_BUS_EVENT,
     SKIP_OBJECTS_BUS_EVENT,
     EXTRUDE_RETRACT_BUS_EVENT,
+    LOAD_FILAMENT_BUS_EVENT,
+    UNLOAD_FILAMENT_BUS_EVENT,
 )
 from .pybambu.commands import (
     PRINT_PROJECT_FILE_TEMPLATE,
     SEND_GCODE_TEMPLATE,
     SKIP_OBJECTS_TEMPLATE,
     EXTRUDER_GCODE,
+    SWITCH_AMS_TEMPLATE,
 )
 
 class BambuDataUpdateCoordinator(DataUpdateCoordinator):
@@ -69,6 +72,8 @@ class BambuDataUpdateCoordinator(DataUpdateCoordinator):
         self.hass.bus.async_listen(SEND_GCODE_BUS_EVENT, self._service_call_send_gcode)
         self.hass.bus.async_listen(SKIP_OBJECTS_BUS_EVENT, self._service_call_skip_objects)
         self.hass.bus.async_listen(EXTRUDE_RETRACT_BUS_EVENT, self._service_call_extrude_retract)
+        self.hass.bus.async_listen(LOAD_FILAMENT_BUS_EVENT, self._service_call_load_filament)
+        self.hass.bus.async_listen(UNLOAD_FILAMENT_BUS_EVENT, self._service_call_unload_filament)
 
     @callback
     def _async_shutdown(self, event: Event) -> None:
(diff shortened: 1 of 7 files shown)


custom_components/bambu_lab/coordinator.py
```diff
@@ -208,11 +213,76 @@ def _service_call_extrude_retract(self, event: Event):
...
         data = event.data
         if not self._service_call_is_for_me(data):
             return
+
+        LOGGER.debug(f"_service_call_load_filament: {data}")
+
+        # Printers with older firmware require a different method to change
+        # filament. For now, only support newer firmware.
+        if not self.get_model().supports_feature(Features.AMS_SWITCH_COMMAND):
+            LOGGER.error(f"Loading filament not yet available for this device")
```
AdrianGarside, line 226: Make this wording clearer that a firmware update is needed since we know it is already available - just not for this specific device until the firmware is updated.

custom_components/bambu_lab/coordinator.py
```diff
@@ -208,11 +213,76 @@ def _service_call_extrude_retract(self, event: Event):
...
+        tray = int(data.get('tray', 1))
+        temperature = int(data.get('temperature', 0))
+
+        if data.get('external_spool') is True:
+            tray = 254
+            # Unless a target temperature override is set, try and find the
+            # midway temperature of the filament set in the ext spool
+            ext_spool = self.get_model().external_spool
+            if data.get('temperature') is None and not ext_spool.empty:
+                temperature = (int(ext_spool.nozzle_temp_min) + int(ext_spool.nozzle_temp_min)) / 2
```
AdrianGarside, line 238: Typo? One should be max?

custom_components/bambu_lab/coordinator.py
```diff
@@ -208,11 +213,76 @@ def _service_call_extrude_retract(self, event: Event):
...
+            # Check the AMS exists and has filament
+            if not self.get_model().ams.data[ams_idx] or self.get_model().ams.data[ams_idx].tray[tray].empty:
+                LOGGER.error(f"AMS tray '{data.get('tray')}' is empty")
+                return False
+
+            # Unless a target temperature override is set, try and find the
+            # midway temperature of the filament set in the ext spool
+            if data.get('temperature') is None:
+                ams_tray = self.get_model().ams.data[ams_idx].tray[tray]
+                temperature = (int(ams_tray.nozzle_temp_min) + int(ams_tray.nozzle_temp_min)) / 2
```
AdrianGarside, line 256: Typo - one should max?

custom_components/bambu_lab/coordinator.py
```diff
@@ -208,11 +213,76 @@ def _service_call_extrude_retract(self, event: Event):
...
+
+    def _service_call_unload_filament(self, event: Event):
+        data = event.data
+        if not self._service_call_is_for_me(data):
+            return
+
+        LOGGER.debug(f"_service_call_unload_filament: {data}")
+
+        if not self.get_model().supports_feature(Features.AMS_SWITCH_COMMAND):
+            LOGGER.error(f"Loading filament not yet available for this device")
```
AdrianGarside, line 274: Clarify that it is available... if you update the firmware.

AdrianGarside (changes requested):
Some fixes seem to be needed and also a merge conflict with your other changes :)

AdrianGarside (approved):

coltography:
Does this only work with AMS or can you load/unload if you just have an external spool?

steveneaston:
@coltography works with both independently. Just provide `external_spool: true` in the load filament action — or use the UI, enable the ext spool option and toggle it on.

GEBones:
Are there any plans to addd this for the H2 series? I have the H2D printer and I don’t see the button or option to add a script for load / unload.

AdrianGarside:
It's been present the whole time, but you have to be in developer lan mode to be able control the printer.

PR #1009, opened 2025-02-15
Update platform support

lunDreame:
- The type of the update entity of the existing binary sensor platform has been changed to an entity belonging to the update platform.
- As bambu lab offline firmware update becomes possible, updates can be implemented by sending payload via MQTT. (https://bambulab.com/en/support/firmware-download/all)
- Additional features that come with it include:
  - Support latest version update (power by: https://github.com/lunDreame/user-bambulab-firmware/blob/main/scripts/fetch.py)
  - Progress when updating
  - Current version 
  - Latest version
  - Latest version compare current version

- Finally, you can update firmware on HomeAssistant locally!
- We still only support the latest version updates and do not support specific version updates.
- It has only been tested on the P1 series, so more test cases for other devices are needed.


lunDreame:
@greghesp , @AdrianGarside 
please review it

AdrianGarside:
@lunDreame thanks for this. I was waiting to get a fresh release out so there was a bit of time to test out this change before it gets into a public release. Our worry is how reliable the firmware update process is.

lunDreame:
Thank you for your response. Does it work well when tested on a personal device? The upgrade dictionary values in the push_all response may not confirm that all printers are unified, so you may have to adjust them separately for a specific printer

AdrianGarside:
I'll have to downgrade a printer firmware to test the upgrade. I am thinking of hiding this new functionality behind the advanced options in the config flow for now. So that users have to explicitly turn it on with a warning it's at their own risk.

lunDreame:
That's right. I tested it with the global P1 series and it works fine. There is no need to distinguish between domestic/global. 

The response to the push_all command upgrade_state is the same, so the upgrade_state may be different for other printers such as A1/Mini and X1, so this can bring about a different upgrade process by adding condition logic to suit the situation. Eventually, we need feedback from users who actually have different printers for this function. 

There is no problem with where each printer is updating normally. It has already been tested through that link (https://github.com/lunDreame/user-bambulab-firmware). However, for the X1 series, the upgrade request payload is different due to security issues. 

It can only be solved if there is a real device, but I don't have that device.

AdrianGarside:
Tested P1S upgrade. Worked well. Only minor issue is that the firmware reported on the device page doesn't update to the new firmware until the integration is reloaded.

Trying to test X1C but it's not detecting that there's an available newer firmware.

lunDreame:
Could you please upload the X1C's push_all response payload? (Response when firmware update is available / Reply when latest firmware) I will check and upload the improved PR :)

AdrianGarside:
The X1 provides it's firmware update data differently:
```
        # Example payload for P1 printer
        # "upgrade_state": {
        #    "sequence_id": 0,
        #    "progress": "100",
        #    "status": "UPGRADE_SUCCESS",
        #    "consistency_request": false,
        #    "dis_state": 1,
        #    "err_code": 0,
        #    "force_upgrade": false,
        #    "message": "0%, 0B/s",
        #    "module": "ota",
        #    "new_version_state": 1,
        #    "cur_state_code": 0,
        #    "new_ver_list": [
        #    {
        #      "name": "ota",
        #      "cur_ver": "01.06.01.02",
        #      "new_ver": "01.07.00.00",
        #      "cur_release_type": 3,
        #      "new_release_type": 3
        #    }
        #
        # Example payload for X1 printer
        # "upgrade_state": {
        #     "ahb_new_version_number": "",
        #     "ams_new_version_number": "",
        #     "consistency_request": false,
        #     "dis_state": 3,
        #     "err_code": 0,
        #     "ext_new_version_number": "",
        #     "force_upgrade": false,
        #     "idx": 4,
        #     "idx1": 3,
        #     "message": "RK1126 start write flash success",
        #     "module": "",
        #     "new_version_state": 1,
        #     "ota_new_version_number": "01.08.02.00",
        #     "progress": "100",
        #     "sequence_id": 0,
        #     "sn": "**REDACTED**",
        #     "status": "UPGRADE_SUCCESS"
        # },
```

lunDreame:
Okay.. I'll try to improve the logic to parse differently. Thx

AdrianGarside:
Fixed that but the upgrade attempt failed. This was the URL from the firmware page and the generated download URL:
```
https://public-cdn.bblmw.com/upgrade/device/offline/BL-P001/01.08.02.00/e750866b34/offline-ota-p001_v01.08.02.00-20240812145251.zip

{
    'user_id': '0',
    'upgrade': 
    {
        'sequence_id': '0',
        'command': 'upgrade_history',
        'src_id': 2,
        'firmware_optional':
        {
            'firmware':
            {
                'version': '01.08.02.00',
                'url': 'https://public-cdn.bblmw.com/upgrade/device/BL-P001/01.08.02.00/product/e750866b34/ota-p001_v01.08.02.00-20240812145251.json.sig',
                'force_update': False,
                'description': '',
                'status': 'release'
            },
            'ams': []
        }
    }
}
```

AdrianGarside:
Trying to view that URL gives an access denied error:

AccessDeniedAccess Denied
This XML file does not appear to have any style information associated with it. The document tree is shown below.
<Error>
<Code>AccessDenied</Code>
<Message>Access Denied</Message>
...
</Error>

lunDreame:
> Fixed that but the upgrade attempt failed. This was the URL from the firmware page and the generated download URL:
> 
> ```
> https://public-cdn.bblmw.com/upgrade/device/offline/BL-P001/01.08.02.00/e750866b34/offline-ota-p001_v01.08.02.00-20240812145251.zip
> 
> {
>     'user_id': '0',
>     'upgrade': 
>     {
>         'sequence_id': '0',
>         'command': 'upgrade_history',
>         'src_id': 2,
>         'firmware_optional':
>         {
>             'firmware':
>             {
>                 'version': '01.08.02.00',
>                 'url': 'https://public-cdn.bblmw.com/upgrade/device/BL-P001/01.08.02.00/product/e750866b34/ota-p001_v01.08.02.00-20240812145251.json.sig',
>                 'force_update': False,
>                 'description': '',
>                 'status': 'release'
>             },
>             'ams': []
>         }
>     }
> }
> ```

See this [url](https://github.com/lunDreame/user-bambulab-firmware/blob/main/assets/x1_series_ams.json) for x1 series.

Also Access DeniedAccess Denied cannot get a proper firmware .sig file if some of the firmware addresses are not correct. For security purposes, some of the addresses timestamp, hash, etc.

lunDreame:
But even if you send the above payload of the X1 series, it won't actually work. 

The X1 series had different firmware commands. I have to find out the actual firmware commands by testing them after parsing, but it's difficult because there's no device. 

Roughly, you can extract commands from the Bambu Handy app logcat MQTT communication log.

PR #1035, opened 2025-02-17
feat(config_flow): add option to disable ssl verification

tarik02:
## Description

Added an option to disable MQTT, FTP and Chamber Image SSL verification which was introduced in https://github.com/greghesp/ha-bambulab/pull/876. It allows to connect to host that looks like Bambu printer, but it is not (for example, proxy that allows more connections than printer).

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Documentation

## Testing

Manual testing with EMQX set up as mqtt bridge:
зображення
зображення

## Additional Notes

https://github.com/disconn3ct/bambu-proxy/issues/2


AdrianGarside:
Really nice addition. I like it being in an expandable advanced section.

PR #1110, opened 2025-03-03
local gcode download feature

Dronedale:
## Description
EDIT: redid my last PR now based on current ha-bambulab release

With this PR I added a feature to be able to download gcode from printer to local HA directory via ftp. The intention is to use this gcode for a custom card for a three.js 3d live gcode preview I am currently working on.

## Type of Change

- [ ] Bug fix
- [x ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Documentation

## Testing

## Additional Notes

to enable this feature I added a config switch to the printer to enable local gcode download (only if ftp is enabled), otherwise it cannot being enabled.
When ftp data is being fetched by '_async_download_task_data_from_printer' it checks if this feature is enabled and if so, it will download the corresponding gcode file to '/config/www/media/ha-bambulab/{self._client._serial}/tmp_gcode/{tmpFile.name}.gcode'

I also added a sensor holding the currently downloaded gcode filename. This is used by my custom card for picking up the right gcode file.

### DIFF
--- a/custom_components/bambu_lab/coordinator.py
+++ b/custom_components/bambu_lab/coordinator.py
@@ -695,6 +695,14 @@ class BambuDataUpdateCoordinator(DataUpdateCoordinator):
     async def set_option_enabled(self, option: Options, enable: bool):
         LOGGER.debug(f"Setting {OPTION_NAME[option]} to {enable}")
         options = dict(self.config_entry.options)
+        
+        if option == Options.DOWNLOAD_GCODE_FILE:
+            if enable:
+                enable = self.get_option_enabled(Options.FTP)
+        if option == Options.FTP:
+            if not enable:
+                options[OPTION_NAME[Options.DOWNLOAD_GCODE_FILE]] = enable
+                
         options[OPTION_NAME[option]] = enable
         self._hass.config_entries.async_update_entry(
             entry=self.config_entry,
@@ -714,6 +722,8 @@ class BambuDataUpdateCoordinator(DataUpdateCoordinator):
             case Options.FTP:
                 force_reload = True
             case Options.TIMELAPSE:
+                force_reload = True
+            case Options.DOWNLOAD_GCODE_FILE:
                 force_reload = True
 
         if force_reload:
--- a/custom_components/bambu_lab/coordinator.py
+++ b/custom_components/bambu_lab/coordinator.py
@@ -723,8 +723,6 @@ class BambuDataUpdateCoordinator(DataUpdateCoordinator):
                 force_reload = True
             case Options.TIMELAPSE:
                 force_reload = True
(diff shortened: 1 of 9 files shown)


custom_components/bambu_lab/switch.py
```diff
@@ -59,6 +59,13 @@
     entity_category=EntityCategory.CONFIG,
 )
 
+DOWNLOAD_GCODE_FILE_SWITCH_DESCRIPTION = SwitchEntityDescription(
+    key="downnload_gcode_file",
```
AdrianGarside, line 63: Typo

custom_components/bambu_lab/coordinator.py
```diff
@@ -715,6 +723,8 @@ async def set_option_enabled(self, option: Options, enable: bool):
                 force_reload = True
             case Options.TIMELAPSE:
                 force_reload = True
+            case Options.DOWNLOAD_GCODE_FILE:
+                force_reload = True
```
AdrianGarside, line 727: The switch doesn't stop existing so I don't think you need to force a reload for this option.

custom_components/bambu_lab/pybambu/models.py
```diff
@@ -1163,6 +1168,24 @@ def _async_download_task_data_from_printer(self):
...
+                                for name in os.listdir(local_gcode_dir):
+                                    path = os.path.join(local_gcode_dir, name)
+                                    if os.path.isfile(path):
+                                        os.remove(path)
+                                with archive.open(f"Metadata/plate_{plate_number}.gcode") as gcode_entry, open(os.path.join(local_gcode_dir, local_gcode_filename), "wb") as target_path:
+                                    shutil.copyfileobj(gcode_entry, target_path)
+                                    self.gcode_file_downloaded = local_gcode_filename
+                            except Exception as e:
+                                self.gcode_file_downloaded = "ERROR"
+                                LOGGER.debug(f"Error while extracting gcode zip entry to target path. {repr(e)}")
```
AdrianGarside, line 1187: This should probably be logged as an actual error.

AdrianGarside (changes requested):

custom_components/bambu_lab/switch.py
Dronedale: applied changes

custom_components/bambu_lab/coordinator.py
Dronedale: applied change

custom_components/bambu_lab/pybambu/models.py
Dronedale: applied change

AdrianGarside (approved):

PR #1126, opened 2025-03-07
fix ui not updating switch state for download gcode file

Dronedale:
## Description

readd force_reload for option download_gcode_file, otherwise the switch state isn't properly updated in ui

## Type of Change

- [x ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Documentation

## Testing

### DIFF
--- a/custom_components/bambu_lab/const.py
+++ b/custom_components/bambu_lab/const.py
@@ -43,6 +43,7 @@ class Options(IntEnum):
     TIMELAPSE = 4,
     MANUALREFRESH = 5,
     FIRMWAREUPDATE = 6,
+    DOWNLOAD_GCODE_FILE = 7
 
 OPTION_NAME = {
     Options.CAMERA:         "enable_camera",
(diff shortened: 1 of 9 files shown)


AdrianGarside (changes requested):
Can you clarify what the problem. The user sets the state in the UI. The integration only sets it on load. I'm not seeing any misbehavior of that switch as-is.

Dronedale:
sure, the problem is when for example the user sets the gcode download switch from off to on (while ftp switch is turned off), the gcode download switch should also visually go back to its off position (because ftp is required to be enabled).
Without force refresh, after enabling the gcode download switch, the switch stays visually in its on position, but is actually off.
Another approach might be to make use of the switch's 'is_on' property from the switch base class which gets called frequently by HA, which would return 'get_option_enabled' for the gcode download option.

Hope this makes it clear, maybe you have a different idea?

AdrianGarside:
Ok. Rather than forcing a refresh which is a hard reload of the integration that is better handled by coordinating the allowed / available states of the switches better.

AdrianGarside:
https://github.com/greghesp/ha-bambulab/pull/1134

Issue #1212, opened 2025-04-08
[Bug] Print preview is huge


MelodyFox-jerrf:
### Home Assistant Version

2025.4.1

### Bambu Lab Integration Version

2.1.11

### Describe the bug

The print preview is very big and blocking everything. I just updated it and this happens: 

![Image](https://github.com/user-attachments/assets/b5e0382d-6895-4835-a52c-6359ab946ad7)

### To Reproduce

1. Insert the print preview card

### Expected Behaviour

To be a normal picture

### What device are you using?

A1 Mini

### Diagnostic Output

```shell
Not Needed
```

### Log Extracts

```shell

```

AdrianGarside:
https://github.com/greghesp/ha-bambulab/releases/tag/v2.1.12

Issue #1208, opened 2025-04-07
[Bug] Graphic Status Card Image Scaling Bug (A1)


VT-14:
### Home Assistant Version

2025.3.3

### Bambu Lab Integration Version

2.1.11

### Describe the bug

![Image](https://github.com/user-attachments/assets/20ef2a1e-ea73-4d0a-bced-b1312c92bdfb)

### To Reproduce

1. Use the Bambu Lab Print Status Card set to the Graphic mode for the A1 ("Simple" mode is scaled correctly).
2. Have a Cover Image to load.
3. (Optional) Laugh.

### Expected Behaviour

To not have a massive Cover Image preventing using the cards around the Status Card.

### What device are you using?

A1

### Diagnostic Output

```shell
{
  "home_assistant": {
    "installation_type": "Home Assistant Container",
    "version": "2025.3.3",
    "dev": false,
    "hassio": false,
    "virtualenv": false,
    "python_version": "3.13.2",
    "docker": true,
    "arch": "x86_64",
    "timezone": "America/New_York",
    "os_name": "Linux",
    "os_version": "6.6.44-production+truenas",
    "run_as_root": true
  },
  "custom_components": {
    "hacs": {
      "documentation": "https://hacs.xyz/docs/use/",
      "version": "2.0.5",
      "requirements": [
        "aiogithubapi>=22.10.1"
      ]
    },
    "bambu_lab": {
      "documentation": "https://github.com/greghesp/ha-bambulab",
      "version": "2.1.11",
      "requirements": [
        "beautifulsoup4",
        "cloudscraper"
      ]
    }
  },
  "integration_manifest": {
    "domain": "bambu_lab",
    "name": "Bambu Lab",
    "codeowners": [
      "greghesp",
      "AdrianGarside"
    ],
    "config_flow": true,
    "dependencies": [
      "device_automation",
      "ffmpeg",
      "frontend",
      "http",
      "lovelace",
      "mqtt",
      "websocket_api"
    ],
    "documentation": "https://github.com/greghesp/ha-bambulab",
    "iot_class": "local_push",
    "issue_tracker": "https://github.com/greghesp/ha-bambulab/issues",
    "requirements": [
      "beautifulsoup4",
      "cloudscraper"
    ],
    "ssdp": [
      {
        "st": "urn:bambulab-com:device:3dprinter:1"
      }
    ],
    "version": "2.1.11",
    "is_built_in": false,
    "overwrites_built_in": false
  },
[...]

PR #1213, opened 2025-04-08
fix cover image height

elax46:
## Description

With the latest version there is a problem with the cover image for a height setting of 250% instead of 50% several users have reported this problem. 

`Cover Image`

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### Link to Issue

some problem requests highlighted by users #1212 #1208 

## Documentation

- [X] No documentation updates were necessary for this change

## Testing

the solution has been tested inside my dev instance of HA


AdrianGarside:
That isn't how the cards updates come in - the come from a separate repo. Editing the generated ts file doesn't work and is unreviewable. And I just fixed this anyway.

elax46:
you should update the contribution guide on this project it's all very confusing and unclear

Issue #1235, opened 2025-04-15
[Feature] AMS 2 Pro - Humidity available as percentage?


Linwood-F:
### Describe the feature

In the Bambu Handy app, the AMS 2 Pro (on a P1S) shows humidity as a percentage, e.g. mine right now says 40%. 

The entity from this device shows a number "Humidity Index" with mine currently showing 2. 

I have no idea if Handy is just translating that index into a guess at humidity, or if it has a percentage available in the data, but if the device is providing a relative percentage it would be nice to get that instead. 

Note the "AMS 2" may be different from the prior AMS's?

### What device is this for?

A1

IetIesAai:
I think I read somewhere the percentage it displays is more or less the same as the index, in that it only updates in 20% increments (eg 1=20% 2=40% 3=60% 4=80% 5= 100% or vice versa) - partially out of interest (I have one ordered): have you seen other numbers then those,  does it also display in smaller/other increments?

Linwood-F:
I think you may be right.  The Handy app last night just showed 40%, this morning 20%, corresponding to indexes 2 and 1.

Over that time (with several containers of Desiccant added) the "real" humidity (non-calibrated but reasonably accurate zooz sensor) went from 45% to 24%. 

So this may not be worth doing.  Or maybe doing the multiplication as Handy does, hard to say if that's more misleading than useful.  I'd probably err on the misleading side.  Feel free to close/ignore this request. 

Re "Have one ordered" - I have a whole 4 days experience with 3D printing, but was rather surprised how easy it was to just plug this in and have it work.  I kept looking for stuff I needed to do, but with the BL RFID's I didn't even have to tell it filament. It can't be used as a dryer (yet), and I'm not sure it makes sense as a dryer if you mix filament types (and expected dryer temps). Have fun!

kurim:
Just as hint, the AMS reports an "humidity_raw" 

![Image](https://github.com/user-attachments/assets/0495953e-3706-4fd6-97e4-7593626b48b0)
Can be seen in this [log](https://github.com/greghesp/ha-bambulab/issues/1240)

This value is also reported by AMS 1, also bambu app change to _raw instead of index

basvdploeg:
> Yep, the percentage is just a percentage based on the humidity index as far as I'm aware.
> 
> I'll keep this open for a bit until your AMS 2 turns up, and to give some time to see if you see anything other than prefixed percentages

[...]

PR #1252, opened 2025-04-23
Add AMS temperature, humidity percentage and drying entities

piitaya:
## Description

- Add AMS temperature entity for P1 (01.07.50.18) (it was already available for X1)
- Add AMS humidity for H2D, P1 (01.07.50.18) and X1, X1C (01.08.50.18). I didn't added X1E because I don't know if it has the right firmware.
- Add drying remaining time for AMS 2 Pro
- Rename `AMS temperature` to `Temperature`. The AMS prefix is not needed because the entity is part of the AMS device.

AMS 2 Pro

CleanShot 2025-04-23 at 23 07 41@2x

AMS 1

CleanShot 2025-04-23 at 17 57 19

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

### Link to Issue

Fixes https://github.com/greghesp/ha-bambulab/issues/1235

## Documentation

## Testing

### DIFF
--- a/custom_components/bambu_lab/pybambu/models.py
+++ b/custom_components/bambu_lab/pybambu/models.py
@@ -182,19 +182,19 @@ class Device:
         elif feature == Features.DOWNLOAD_GCODE_FILE:
             return True
         elif feature == Features.AMS_HUMIDITY:
-            if (self.info.device_type == "H2D"):
+            if (self.info.device_type == Printers.H2D):
                 return True
-            elif (self.info.device_type == "X1" or self.info.device_type == "X1C") and self.supports_sw_version("01.08.50.18"):
+            elif (self.info.device_type == Printers.H2D or self.info.device_type == Printers.X1C) and self.supports_sw_version("01.08.50.18"):
                 return True
-            elif (self.info.device_type == "P1S" or self.info.device_type == "P1P") and self.supports_sw_version("01.07.50.18"):
+            elif (self.info.device_type == Printers.P1S or self.info.device_type == Printers.P1P) and self.supports_sw_version("01.07.50.18"):
                 return True
             return False
         elif feature == Features.AMS_DRYING:
-            if (self.info.device_type == "H2D"):
+            if (self.info.device_type == Printers.H2D):
                 return True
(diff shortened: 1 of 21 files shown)


custom_components/bambu_lab/pybambu/models.py
```diff
@@ -177,7 +181,22 @@ def supports_feature(self, feature):
             return False
         elif feature == Features.DOWNLOAD_GCODE_FILE:
             return True
-
+        elif feature == Features.AMS_HUMIDITY:
+            if (self.info.device_type == Printers.H2D):
+                return True
+            elif (self.info.device_type == Printers.H2D or self.info.device_type == Printers.X1C) and self.supports_sw_version("01.08.50.18"):
```
AdrianGarside, line 187: H2D check here is duplicative with lines 184-185 and the software version is unlikely to be correct.

custom_components/bambu_lab/pybambu/models.py
```diff
@@ -177,7 +181,22 @@ def supports_feature(self, feature):
...
+                return True
+            elif (self.info.device_type == Printers.P1S or self.info.device_type == Printers.P1P) and self.supports_sw_version("01.07.50.18"):
+                return True
+            return False
+        elif feature == Features.AMS_DRYING:
+            if (self.info.device_type == Printers.H2D):
+                return True
+            elif (self.info.device_type == Printers.H2D or self.info.device_type == Printers.X1C) and self.supports_sw_version("01.08.50.18"):
```
AdrianGarside, line 195: H2D incorrect here too.

custom_components/bambu_lab/translations/es.json
```diff
@@ -446,9 +446,6 @@
       "humidity_index": {
         "name": "Índice de humedad"
       },
-      "ams_temp": {
```
AdrianGarside, line 449: Removing this looks like it will break the sensor name as it's still referenced as a translation key.

AdrianGarside: And also breaks the cards since we the translation key is what we need to identify the sensors since the unique key isn't synchronously accessible.

PR #1264, opened 2025-04-27
Fix AMS2 updates

AdrianGarside:
## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Documentation

## Testing


custom_components/bambu_lab/definitions.py
```diff
@@ -499,6 +499,7 @@ class BambuLabBinarySensorEntityDescription(BinarySensorEntityDescription, Bambu
     ),
     BambuLabAMSSensorEntityDescription(
         key="humidity",
+        translation_key="humidity",
```
piitaya, line 502: Humidity does not need to be translated as it's already use the translation from device class

custom_components/bambu_lab/pybambu/models.py
```diff
@@ -184,15 +184,15 @@ def supports_feature(self, feature):
...
+            elif (self.info.device_type == Printers.X1C) and self.supports_sw_version("01.08.50.18"):
                 return True
             elif (self.info.device_type == Printers.P1S or self.info.device_type == Printers.P1P) and self.supports_sw_version("01.07.50.18"):
                 return True
             return False
         elif feature == Features.AMS_DRYING:
             if (self.info.device_type == Printers.H2D):
                 return True
-            elif (self.info.device_type == Printers.H2D or self.info.device_type == Printers.X1C) and self.supports_sw_version("01.08.50.18"):
+            elif (self.info.device_type == Printers.X1C) and self.supports_sw_version("01.08.50.18"):
```
piitaya, line 195: Printers.X1 is missing. I think I set H2B in my PR but it should be X1

custom_components/bambu_lab/pybambu/models.py
```diff
@@ -1683,7 +1683,6 @@ class AMSInstance:
     serial: str
     sw_version: str
     hw_version: str
-    model: str
```
piitaya, line 1686: Why is this removed? It is needed to be able to do `coordinator.get_model().ams.data[index].model == "AMS 2 Pro"` in the definition.py right?

custom_components/bambu_lab/definitions.py
AdrianGarside: We need the translation_key set because we use that to unique identify the sensors in the cards because getting the actual unique_id can't be done synchronously so we use translation key as a proxy for it.

custom_components/bambu_lab/pybambu/models.py
AdrianGarside: Ah, that's what was going on :)

custom_components/bambu_lab/definitions.py
piitaya: I suspected that was the reason. It feels like a bit hacky to do that. The card can also select the right sensor using the device class (device class : humidity and no translation_key)

custom_components/bambu_lab/pybambu/models.py
piitaya: Yeah probably a mistake I did with I rebased my PR to use the enum.

PR #1285, opened 2025-05-12
Block control sensors to match Bambu restrictions

AdrianGarside:
## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Documentation

## Testing


Skydev0h:
H2D _does_ have Developer mode

AdrianGarside:
Thanks for letting me know that. I’d been told it didn’t.

Skydev0h:
> Thanks for letting me know that. I’d been told it didn’t. Has it had that since it first dropped or did a firmware update add it?

I do not know because I updated the firmware right away and did not check it beforehand.

But I will be able to check Laser edition once it arrives (but thats no earlier than in a week) it it would have older firmware.

Skydev0h:
@AdrianGarside unfortunately I can confirm that out of the box H2D Laser with FW version 01.00.04.18 does NOT have "Developer Mode" option.

image

image

On the other hand, H2D with FW version 01.01.01.00 does have the option (see screenshots in the previous reply).

Therefore, I can assume, that it is present on FW version 01.01 and is not present on version 01.00.

PR #1436, opened 2025-07-05
Fix door sensor

### LINKED ISSUE #1421: [Bug] H2D enclosure door sensor not working
### Home Assistant Version

2025.6.3

v2.1.23

### Describe the bug

Opening or closing the enclosure door has no effect on the state of the "Enclosure door" sensor.

A bit of MQTT capture indicates that the door status is likely in the `print.stat` field. This changes when the door is opened/closed, but `print.home_flag` does not:
* Example closed: ` "stat": "46258000",`
* Example opened:  "stat": "46A58000",`

This would align with the bitmask for DOOR_OPEN, just a different field needs to be checked.

### To Reproduce

1. Go to any place the enclosure door sensor state is displayed (in settings, a dashboard, developer tools, etc.)
2. Open the enclosure door on the printer.
3. Close the enclosure door.

### Expected Behaviour

When the enclosure door is opened, the sensor state should be 'open' ('on').

H2D

### Diagnostic Output

```shell
Available on request
```

### Log Extracts

```shell

```

### Other Information

https://forum.bambulab.com/t/mqtt-report-of-door-sensor-state/59955 indicates this isn't working for at least one X1C user either. I'm not seeing any references to this bit in the BambuStudio sources, just a few issue reports about door status warnings being noisy.

I'd be happy to work on this or any of the other issues I've opened. It's a small way to give back for such a useful integration.

pkolbus:
## Description

As part of the work for H2D support, Bambu Lab seems to have reorganized some of the status flags. The door open flag is in `stat` in the MQTT report message, based on differential analysis rather than any published source.

Add a dataclass to parse `stat` and favor this on the H2D. For the X1/X1C, I'm assuming that this moved at some point via firmware update. 1.09.00.00 is a guess based the [X1/X1C release notes](https://wiki.bambulab.com/en/x1/manual/X1-X1C-AMS-firmware-release-history) -- this version updates the user interface, adds support for AMS 2 Pro/AMS HT, and has a recommendation to update Bambu Studio to 2.0.3.54.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### Link to Issue

Fixes #1421.

## Documentation

## Testing

Manually verified with my H2D. I don't have an X1 series printer. Existing unit tests in pybambu/tests fail when I try to run them (Ubuntu 24.04, Python 3.12), and the run_tests.sh script seems meant only for adrian to run. If these tests are supposed to work I'd be happy to dig deeper.

## Additional Notes

Uncertain of the right design for this change -- would it be better to abstract the flags in pybambu to simplify the integration and any other callers, or is the model intended to follow the MQTT message schema?

I don't have access to an X1 series printer, so I'm also uncertain about the specific X1 version that might have changed this, or if this was always broken. The BambuStudio changes that make use of stat are made in June/July 2024 but committed in 2025 for the 2.0 software. [A message by hyperkick in the Bambu Lab Forums](https://forum.bambulab.com/t/mqtt-report-of-door-sensor-state/59955/4) implies that `home_flag` was incorrect for X1 series printers as of January 2025, but hyperkick may have been using prerelease firmware.  

There are other documented (used by BambuStudio) flags in stat and other new fields. I've tried to keep this PR focused on the bug, but happy to make this more complete in this PR or another.

### DIFF
--- a/custom_components/bambu_lab/definitions.py
+++ b/custom_components/bambu_lab/definitions.py
@@ -159,8 +159,8 @@ PRINTER_BINARY_SENSORS: tuple[BambuLabBinarySensorEntityDescription, ...] = (
         translation_key="door_open",
         device_class=BinarySensorDeviceClass.DOOR,
         entity_category=EntityCategory.DIAGNOSTIC,
-        available_fn=lambda self: self.coordinator.get_model().home_flag.door_open_available,
-        is_on_fn=lambda self: self.coordinator.get_model().home_flag.door_open,
+        available_fn=lambda self: self.coordinator.get_model().stat_flag.door_open_available or self.coordinator.get_model().home_flag.door_open_available,
+        is_on_fn=lambda self: self.coordinator.get_model().stat_flag.door_open if self.coordinator.get_model().stat_flag.door_open_available else self.coordinator.get_model().home_flag.door_open_available,
         exists_fn=lambda coordinator: coordinator.get_model().supports_feature(Features.DOOR_SENSOR),
     ),
     BambuLabBinarySensorEntityDescription(
--- a/custom_components/bambu_lab/pybambu/const.py
+++ b/custom_components/bambu_lab/pybambu/const.py
@@ -209,6 +209,11 @@ class Home_Flag_Values(IntEnum):
     SUPPORTED_PLUS                      = 0x08000000,
     # Gap
 
+class Stat_Flag_Values(IntEnum):
+    # Gap
+    DOOR_OPEN                           = 0x00800000,
+    # Gap
+
 class BambuUrl(IntEnum):
     LOGIN = 1,
     TFA_LOGIN = 2,
--- a/custom_components/bambu_lab/pybambu/models.py
+++ b/custom_components/bambu_lab/pybambu/models.py
@@ -45,6 +45,7 @@ from .const import (
     Features,
     FansEnum,
     Home_Flag_Values,
+    Stat_Flag_Values,
     Printers,
     SPEED_PROFILE,
     GCODE_STATE_OPTIONS,
@@ -78,6 +79,7 @@ class Device:
         self.print_error = PrintError(client = client)
         self.camera = Camera(client = client)
         self.home_flag = HomeFlag(client=client)
+        self.stat_flag = StatusFlag(client=client)
         self.extruder = Extruder(client=client)
         self.extruder_tool = ExtruderTool(client=client)
         self.push_all_data = None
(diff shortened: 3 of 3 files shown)


AdrianGarside:
This should be abstracted in the pybambu layer so that callers don't need to care what the printer model is. The HomeFlag class was added a long time ago but really it should have just been part of the Info class rather than its own class. And then this new source of door flag state would just be hidden inside that.

AdrianGarside:
To clarify. I don't object to the HomeFlag class existing - just its exposure as part of the pybambu API surface. HomeFlag is a data concept, not a physical one. So if you wanted to tweak your PR to move the door open status to be a new property on the Info class (and have it do the indirect to your new status class or to the existing homeflag class as appropriate), that would be fine for now. 

As for the X1 they seem to have aligned it to the H2D payload as of the public 1.9 firmware release. I am using the existence of new data in the mqtt to decide when to parse the new format vs the old (which usually still exists). That's working ok so far but with the mqtt payload being a delta in some cases that may not work in all scenarios. And it may make it harder to leave the existing structure largely in place as I suggested above but I think either being enabled will be correct and work for now until I do a bigger refactor.

pkolbus:
Thanks for the review and tips. I’ll keep working on this as suggested.

miguelangellv:
Any news about this? I want make a automation that turn on my desk light when I open my X1C's door

pkolbus:
Apologies for the long delay, life has gotten in the way and doing this right has prompted actually setting up a devcontainer. Have that bit done recently, and I’m planning to get this done this weekend.

pleasantone:
I can confirm that the current main branch code (not this pull request) works fine for the X1C with both old and new firmware.

miguelangellv:
> I can confirm that the current main branch code (not this pull request) works fine for the X1C with both old and new firmware.

I have the last firmware and developer mode.

I check it now and the door sensor does not works properly. If I open the door is not detected, but I have one open detection today, but I open the door about 10-15 times...
 
 I don't know when is detected and when is not

pleasantone:
Works perfectly for me, printer is currently NOT in developer mode. 01.10.00.00 firmware. Tested 10/10 times.

pkolbus:
(Finally) updated and ready for review. Rebased to resolve the merge conflict (and because it's been so long) and added unit tests. Ended up removing door_open from HomeFlag and doing all the work in the Info class rather than keeping parsing in separate data field objects; this provides a clearer API, cohesive implementation, and simplifies unit tests.

AdrianGarside:
That's much cleaner. Thank you for addressing my feedback.

Issue #1447, opened 2025-07-11
[Feature] Being able to observe and control some H2D features (+maybe found reliable dev mode flag)


Skydev0h:
### Describe the feature

There are some features that may be useful to be able to be observed and / or controlled in H2D.
These include:
* Heatbed LED (easy)
* Sound option (observing easy, modifying signed) - may be super useful to remotely control Sound that is otherwise impossible and to automate it (for example, disable sound at evening and night time)
* Video recording mode (and resolution, may be relevant not only to H2D, easy)
* Also, anti-fire buzzer - see comment below (observing easy, modifying signed)

More detailed information about observing and modifying listed values is available in Other Information.

P.S. Likely, I have found a reliable Dev mode flag, please see below for more details.

### What device is this for?

H2D

### Other Information

**Heatbed LED**
Observed (note the led_mode value):
```
{"system":{"command":"ledctrl","interval_time":0,"led_mode":"off","led_node":"heatbed_light","led_off_time":0,"led_on_time":0,"loop_times":0,"reason":"did not find the valid led: heatbed_light","result":"fail","sequence_id":"0"}}
{"system":{"command":"ledctrl","interval_time":0,"led_mode":"on","led_node":"heatbed_light","led_off_time":0,"led_on_time":0,"loop_times":0,"reason":"did not find the valid led: heatbed_light","result":"fail","sequence_id":"0"}}
```

Modified (no restrictions):
```
{
  "system": {
    "sequence_id": "0",
    "command": "ledctrl",
    "led_node": "heatbed_light",
    "led_mode": "off",
    "led_on_time": 0,
    "led_off_time": 0,
    "loop_times": 0,
    "interval_time": 0
  }
}

{
  "system": {
    "sequence_id": "0",
    "command": "ledctrl",
    "led_node": "heatbed_light",
    "led_mode": "on",
    "led_on_time": 0,
    "led_off_time": 0,
    "loop_times": 0,
    "interval_time": 0
  }
}
```

**Sound**
Observed:
```
{"print":{"command":"print_option","result":"success","sequence_id":"78","sound_enable":true}}
{"print":{"command":"print_option","result":"success","sequence_id":"80","sound_enable":false}}
```

Modified (requires Developer mode / signing code or MQTT signing proxy - makes sense to disable editing if "mqtt encryption enabled"):
```
{
  "print": {
    "sequence_id": "0",
    "command": "print_option",
    "sound_enable": false
  }
}

{
  "print": {
[...]

PR #1449, opened 2025-07-12
Fixed MQTT security detection, added H2D sound toggle, buzzer buttons, heatbed LED

Skydev0h:
## Description

This PR contains the following changes and bug fixes:
* Fixed developer mode detection, or, actually, "MQTT security mode flag" - that is, enforcement of signature for "print"-class commands. It now properly works even in real time (especially actual when attaching module, and it gets temporarily disabled automatically).
  * The "Developer mode" option in config becomes unused and should be carefully removed.
  * MQTT encryption firmware detection MAY also be not needed anymore - it is possible to reliably determine whether the printer requires signature or not - actually, Bambu Studio seems to rely on this flag - once Dev mode is deactivated, it immediately configures security certificate and starts signing messages, once Dev mode is activated it immediately stops signing them.
* Fixed sound configuration entity not appearing for H2D - in fact, xcam and toggle logic is identical to A1.
* Fixed activated sound configuration entity icon - it was pointing to non existent one.
* Added buttons to activate or silence H2D buzzer / siren.
  * Because it is not possible to properly obtain it's state, and it is quite not stateful (for example, beeping mode only beeps for about 10 seconds), it is implemented as buttons rather than switch. Tri-state nature of this variable (and, possibly, more in the future) also contribute to this.
* Added heatbed light control for the H2D.
  * This one is quite finnicky, everything works perfectly, however, it is not possible to obtain it's state due to bug in firmware. It is only possible to know our own changes, or observe changes by monitoring responses to commands (including printer UI), which I implemented.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

### Link to Issue

https://github.com/greghesp/ha-bambulab/issues/1447

## Documentation

- [x] I have updated the relevant documentation to reflect these changes

## Testing

- [x] Tested on real devices (H2D, X1C, P1S)

N.B. Test script is not quite functional, it anchors to a specific absolute directory.
In theory tests may fail if they did not correctly replicate "fun" value in accordance to Developer mode setting.

## Additional Notes

"Developer mode" configuration option is not needed anymore after these changes, because now the state is properly tracked by analyzing information that is sent by the printer. Nevertheless, I was not sure how to correctly remove it without breaking anything - that is highly recommended to do.

### DIFF
--- a/custom_components/bambu_lab/pybambu/models.py
+++ b/custom_components/bambu_lab/pybambu/models.py
@@ -58,7 +58,7 @@ from .commands import (
     CHAMBER_LIGHT_2_OFF,
     PROMPT_SOUND_ENABLE,
     PROMPT_SOUND_DISABLE,
-    SPEED_PROFILE_TEMPLATE,
+    SPEED_PROFILE_TEMPLATE, BUZZER_SET_SILENT, BUZZER_SET_ALARM, BUZZER_SET_BEEPING,
 )
 
 class Device:
@@ -250,6 +250,8 @@ class Device:
             return False
         elif feature == Features.MQTT_ENCRYPTION_ENABLED:
             return self.print_fun.mqtt_signature_required()
+        elif feature == Features.FIRE_ALARM_BUZZER:
+            return (self.info.device_type == Printers.H2D)
         return False
     
     def supports_sw_version(self, version: str) -> bool:
@@ -1833,6 +1835,16 @@ class Info:
             self._client.publish(PROMPT_SOUND_ENABLE)
         else:
             self._client.publish(PROMPT_SOUND_DISABLE)
+
+    def buzzer_silence(self):
+        self._client.publish(BUZZER_SET_SILENT)
+
+    def buzzer_fire_alarm(self):
+        self._client.publish(BUZZER_SET_ALARM)
+
+    def buzzer_attention_beep(self):
+        self._client.publish(BUZZER_SET_SILENT) # need to reset first for it to work properly
+        self._client.publish(BUZZER_SET_BEEPING)
        
 
 @dataclass
--- a/custom_components/bambu_lab/pybambu/models.py
+++ b/custom_components/bambu_lab/pybambu/models.py
@@ -49,7 +49,7 @@ from .const import (
     SPEED_PROFILE,
     GCODE_STATE_OPTIONS,
     PRINT_TYPE_OPTIONS,
-    TempEnum,
+    TempEnum, Print_Fun_Values,
 )
 from .commands import (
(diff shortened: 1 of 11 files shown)


custom_components/bambu_lab/pybambu/models.py
```diff
@@ -2689,6 +2736,29 @@ def p1s_upgrade_installed(self) -> bool:
...
+        self._encryption_enabled = False
+        self._int_value: int = 0
+
+    def print_update(self, data: dict) -> bool:
+        old_data = f"{self.__dict__}"
+        self._value = data.get("fun", str(self._value))
+        self._int_value = int(self._value, 16) if self._value else 0
+        return (old_data != f"{self.__dict__}")
+
+    def mqtt_signature_required(self) -> bool:
```
AdrianGarside, line 2758: This should be a property

custom_components/bambu_lab/pybambu/models.py
```diff
@@ -266,12 +275,14 @@ class Lights:
...
     chamber_light_override: str
     chamber_light2_override: str
+    heatbed_light: str
     work_light: str
 
     def __init__(self, client):
         self._client = client
         self.chamber_light = "unknown"
         self.chamber_light2 = "unknown"
+        self.heatbed_light = "unknown"
```
AdrianGarside, line 285: For the other lights there's a significant delay between asking for a light state change and getting the mqtt payload that it's been applied. The override variants track a brief override state where we assume the command we sent worked to give a better user experience vs seeing the light go on, then briefly go back to off, before finally going back to on again in home assistant. Do we need that for this one too?

AdrianGarside (approved):

custom_components/bambu_lab/pybambu/models.py
Skydev0h: Not really, because we cannot get the state of this light due to a bug in printer's firmware (described in comments).
Currently it functions entirely as "override" by tracking the state locally, and reacting to internal "MQTT" message generated by UI when state is toggled (or remembering what we done).
Maybe it will be reasonable to add it later when this light will be actually tracked in firmware.

custom_components/bambu_lab/pybambu/models.py
Skydev0h: Oops, a little oversight on my part :)
Should not be too hard to adjust it later, currently this function (property) is only used in one place.

AdrianGarside: Yep. Easy to fix so at any point so I didn't want to block on it.

Issue #1545, opened 2025-09-28
H2D HMS Error Code HMS_0300_A700_0003_0001


Chompy80:
### Describe the feature

A common error code on the H2D is HMS_0300_A700_0003_0001, however the integration does not provide a description for this error.  The printer gives the description below, if you're interested in adding it to a future release.

HMS_0300_A700_0003_0001

<img width="808" height="76" alt="Image" src="https://github.com/user-attachments/assets/7693cae2-e27d-480b-9e28-edb69b7d5fdb" />

### What device is this for?

H2D

pkolbus:
It looks like Bambu Studio changed their HMS resources again -- there are now three different device ID types (093, 094, 239). The  scripts in this repo are only processing the 094, new HMS codes like this one and the new maintenance reminders are in 093. Sometimes the message text is different between these; e.g. in 094 the message references the "nozzle" but in 093 the text says "right nozzle".

@AdrianGarside what do you think about approaches here? It wouldn't be too hard for update_const.py to also download and process the 093 and 239, but Bambu Studio also has translations of these HMS messages... wondering if it would be better to mirror all of the files into this repository, or have the integration itself fetch the needed json file(s) from github based on the printer device(s) (and language?). Happy to work on the PR as I've got one of the maintenance reminder codes currently.

AdrianGarside:
I’d discovered that myself and was wondering if a merge approach would work but then we need to decide how to handle the duplicates - which wins? If they are device specific that explains some of my confusion. I added a live load of resources to the integration as a backup at one point but it added complexity and wouldn’t work for anyone who blocks their home assistant accessing arbitrary web endpoints. I’d rather update the retrieval script to pull everything locally for checkin to the repo.

AdrianGarside:
Have you worked out how the numbers relate to models? 094/239 seem to have left/right extruder mentioned while 093 doesn't. So 093 seems to be P1/X1/A1 maybe. But not sure the difference between 094 and 239.

AdrianGarside:
293 seems to be solely additive plus some level of different text but meaning the same thing. I don't see any removed entries vs 094.

So I guess 093 for single nozzle devices, and 239 for dual nozzle until someone tells me there's a better choice.

pkolbus:
Great question. I'd actually been looking at this today to make sure unknowns are handled well.

[...]

PR #1566, opened 2025-10-12
Update HMS and print error strings to be printer- and language specific

pkolbus:
## Description

Update HMS and print error strings to be printer- and language specific.

Descriptions for HMS and printer error code can vary based on the printer; for example errors about "the nozzle" on X1 are for "the right nozzle" on H2D. Additionally, translations of the message text are available in a number of languages, and the best source of HMS messages is once again e.bambulab.com (when queried properly); Bambu Studio sources on GitHub are missing new H2D messages for printer maintenance.
* Replace scripts/update_const.py with update_error_text.py; this queries e.bambulab.com using the same parameters as Bambu Studio to fetch the text for all known printer models and languages. In the query, printer models are identified by serial number prefix (mapping based on https://wiki.bambulab.com/en/general/find-sn). The languages queried are those that Bambu Studio or this integration have been translated into. The error strings are written to a JSON file per printer and per language.
* Update get_HMS_error_text() and get_print_error_text() to search for a suitable string given an error code, device type, and user language. First preference is to match the user language. If the printer is unknown or the message is not defined for the printer, the H2D strings are used. JSON files are loaded on demand, and the functools.cache decorator is used to improve performance for subsequent lookups.
* Add unit tests for the error lookup functions specifically, as well as use in HMSList and PrintError.
* Remove obsolete files.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

### Link to Issue

#1545

## Documentation

## Testing

Added unit tests, verified on my production HomeAssistant and H2D (has one HMS error) including changing system language between English and Spanish.

## Additional Notes

The H2S prefix of 093 is based on features implied by H2D and print messages. laser/cutter references, no distinction between left/right nozzle in messages, and an error about the 40W laser being unsupported.
The H2D Pro has a serial number prefix of 239. It looks like support for that printer is missing in general, so I left that out.

A possible future improvement might be to machine-translate the missing languages with Google, as is done for other strings.

### DIFF
--- a/custom_components/bambu_lab/pybambu/hms_error_text/hms_H2D_en.json
+++ b/custom_components/bambu_lab/pybambu/hms_error_text/hms_H2D_en.json
@@ -1146,7 +1146,7 @@
     "0500070000020101": "unknown",
     "0500070000020201": "unknown",
     "0500400000010039": "Laser module Serial Number error",
-    "0500400000010040": "Cutting module Serial Number error",
+    "0500400000010040": "Cutting module Serial Number error.",
     "0501040000010044": "The firmware of AMS B does not match the printer. Please update it on the 'Firmware' page.",
     "0501040000030002": "Threaded rods need lubrication now.",
     "0501040000030004": "Please clean and lubricate the X-axis linear rail, YZ-axis linear rods, and the extruder lift linear rail.",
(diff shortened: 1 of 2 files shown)


AdrianGarside:
@pkolbus this is awesome.

There is one problem that I don't have a solution for - the user language coming from hass.config.language isn't really the user language. It's the overall language of the home assistant. For most usage it'll still match the user but in scenarios where the user had selected a different UI language to the base home assistant install, we won't be able to respect that here. I plumbed that language through before I understood what I was trying to do wasn't truly possible.

The difficulty with home assistant is that the integration don't run as a user, they're just data sources. UI like cards does and can get the actual user language. And so home assistant does the translation automatically on behalf of an integration if you align to the handful of self-inconsistent ways to hand off the strings.

PR #1771, opened 2025-12-09
Fix spool label indenting

tgmatt:
## Description

This is a minor change, but it fixes the spool labelling indentation on the AMS cards. They're out by a small amount, but it's enough to be irritating to OCD-types like me.

I hope this is the correct way of fixing this issue, as I couldn't see any other way to modify these values.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### Link to Issue

N/A

## Documentation

## Testing

## Additional Notes

Before changes:
image

After changes:
image


AdrianGarside:
@tgmatt The card javascript comes from a separate github repo:
https://github.com/greghesp/ha-bambulab-cards

If you let me know the change you made, I can apply it there. The minified js in this repo is too large and unreadable to diff to work out what change you made.

AdrianGarside:
It may be that the change you made won't work as-is because there are two different AMS images and they don't align the same way. The labels are currently correctly positioned on the AMS 1 image. So if all you did was adjust them to match the AM 2 image, that just breaks the AMS 1 layout instead.

tgmatt:
Ahh, fair enough. I adjusted the value of the `left` CSS for each of the spools. I guess they'll need to be configured separately for both types of AMS.

These are the changes I made, for reference:
```
.spool-1 { left: 14% }
.spool-2 { left: 33.3% }
.spool-3 { left: 52.4% }
.spool-4 { left: 71.5% }
```

EDIT: To be clear, these were made to centre them correctly for the AMS 2. It seems the AMS 1 is also not centered perfectly, but that'll still need different values to the AMS 2 values.

Issue #972, opened 2025-02-11
[Bug] Start Time unavailable if HA restarted during print


MustardCat:
### Home Assistant Version

2025.2.2

### Bambu Lab Integration Version

V2.1.0

### Describe the bug

If Home Assistant is restarted during a print, sensor.ZZZ_start_time will become unavailable.

Manually setting the state via Developer Tools will temporarily update it but seconds later it'll revert back to "unavailable"

### To Reproduce

1. Start a print
2. Check the sensor.ZZZ_start_time entity
3. Verify entity has a valid start time
4. While printing, reboot Home Assistant
5. Check the sensor.ZZZ_start_time entity
6. Verify entity is now unavailable

### Expected Behaviour

Entity will retain start time information upon reboot

### What device are you using?

P1S

### Diagnostic Output

```shell
{
  "home_assistant": {
    "installation_type": "Home Assistant OS",
    "version": "2025.2.2",
    "dev": false,
    "hassio": true,
    "virtualenv": false,
    "python_version": "3.13.1",
    "docker": true,
    "arch": "aarch64",
    "timezone": "America/Los_Angeles",
    "os_name": "Linux",
    "os_version": "6.6.62-haos-raspi",
    "supervisor": "2025.02.0",
    "host_os": "Home Assistant OS 14.2",
    "docker_version": "27.2.0",
    "chassis": "embedded",
    "run_as_root": true
  },
  "custom_components": {
    "smartthinq_sensors": {
      "documentation": "https://github.com/ollo69/ha-smartthinq-sensors",
      "version": "0.41.1",
      "requirements": [
        "pycountry>=23.12.11",
        "xmltodict>=0.13.0",
        "charset_normalizer>=3.2.0"
      ]
    },
    "presence_simulation": {
      "documentation": "https://github.com/slashback100/presence_simulation",
      "version": "4.12",
      "requirements": []
    },
    "hacs": {
      "documentation": "https://hacs.xyz/docs/use/",
      "version": "2.0.5",
      "requirements": [
        "aiogithubapi>=22.10.1"
      ]
    },
    "spotcast": {
      "documentation": "https://github.com/fondberg/spotcast",
      "version": "v4.0.0",
      "requirements": [
        "spotipy==2.23.0"
      ]
    },
    "bambu_lab": {
      "documentation": "https://github.com/greghesp/ha-bambulab",
      "version": "2.1.0",
      "requirements": [
        "cloudscraper"
      ]
    },
[...]

PR #1838, opened 2026-01-12
LAN Only: Persist start_time state across reboots using a RestoreEntity mixin and allow for total_usage calculation.

baudneo:
## Description

Fixes the 'Start Time' entity losing its state when HASS reboots and running in LAN only mode. Basically, it uses the HASS `RestoreEntity` to persist its state across reboots. The code is backwards compatible and I hope follows the rest of the code structure accordingly.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### Link to Issue

https://github.com/greghesp/ha-bambulab/issues/972

## Documentation

## Testing

I didn't see in the [contributing guide](https://docs.page/greghesp/ha-bambulab/misc/contributing) how you want tests done. Seems that mainly applies to the frontend code. This is a minimal change to the python code (2 files) and shouldn't affect any testing AFAIK.

## Additional Notes

I tested by rebooting while a print was running and it works. No errors in logs and the other sensors work as they did before.

### DIFF
--- a/custom_components/bambu_lab/sensor.py
+++ b/custom_components/bambu_lab/sensor.py
@@ -134,7 +134,7 @@ class BambuLabRestoreSensor(BambuLabSensor, RestoreEntity):
                     if dt_value.tzinfo is None:
                         dt_value = dt_value.replace(tzinfo=dt_util.DEFAULT_TIME_ZONE)
                     job.start_time = dt_value
-                    LOGGER.info(f"LAN Mode: Injected restored start_time into pybambu: {dt_value}")
+                    LOGGER.debug(f"LAN Mode: Injected restored start_time into pybambu: {dt_value}")
                 except (ValueError, TypeError):
                     LOGGER.error("Failed to parse restored start_time string for pybambu")
                     return None


baudneo:
Spoke too soon, it's intermittently keeping it's state. Might be when the coordinator does its first data pull. I'll debug farther after work this evening.

Edit: almost there. I confirmed the state is saving and it pulls the saved state on init. Due to the availability_fn and value_fn lambda callbacks, I'll need to implement the restore state in pybambulab. It seems pybambulab is only used in this add-on, so using it to do the restore should be ok, rather than in the integration code.

```
2026-01-12 21:46:59.677 INFO (MainThread) [custom_components.bambu_lab] initializing RestoreSensor...
2026-01-12 21:46:59.677 INFO (MainThread) [custom_components.bambu_lab] State pulled from Restore Entity: <state sensor.p1s_01XXXXXXX_start_time=2026-01-12 20:04:00; friendly_name=R2D2 P1S Start time @ 2026-01-12T20:04:01.334538-07:00>
```

baudneo:
Alright, I think I've got it handled. It also updates `self.coordinator.get_model().print_job.start_time` so it can be used in the total usage hours calcs.

```
2026-01-13 21:47:09.017 INFO (MainThread) [custom_components.bambu_lab] LAN Mode: Injected restored start_time into pybambu: 2026-01-13 21:18:00-07:00
```

Proof of total usage being incremented at end after the start_time restoration in between start and finish of print:

Issue #1841, opened 2026-01-13
[Bug] Unload filament action broken


Scabattoir:
### Home Assistant Version

2026.1.1

### Bambu Lab Integration Version

v2.2.19

### Describe the bug

The unload filament action is not working for the last couple months (about 3-4-5 last versions) on my P1S.
Tried both the external spool and the AMS and both seem to be dysfunctional now.
It was working all fine before that.
Firmware: 01.07.00.00

### To Reproduce

Call the action, see no effect.

### Expected Behaviour

Call the action, see the printer unload the filament.

### What device are you using?

P1S

### Diagnostic Output

[config_entry-bambu_lab-01JZB4QCYGDQVGKN8390WR3EC2.json](https://github.com/user-attachments/files/24595845/config_entry-bambu_lab-01JZB4QCYGDQVGKN8390WR3EC2.json)

### Log Extracts

Logger: custom_components.bambu_lab
Source: custom_components/bambu_lab/coordinator.py:404
integration: Bambu Lab ([documentation](https://github.com/greghesp/ha-bambulab), [issues](https://github.com/greghesp/ha-bambulab/issues))
First occurred: 17:43:20 (9 occurrences)
Last logged: 18:04:22

Invalid data payload, missing entity_id: {'device_id': 'f15a73c566b3b81e5e60df5d51f16b01', 'service': 'unload_filament'}
Invalid data payload, missing entity_id: {'device_id': '2a5fee85b13dec21e8fb5442f1973068', 'service': 'unload_filament'}
Invalid data payload, missing entity_id: {'device_id': '614dab3c8f942abe321bd41210aa0798', 'service': 'unload_filament'}

### then:

Logger: homeassistant
Source: custom_components/bambu_lab/coordinator.py:560
integration: Bambu Lab ([documentation](https://github.com/greghesp/ha-bambulab), [issues](https://github.com/greghesp/ha-bambulab/issues))
First occurred: 17:43:39 (9 occurrences)
Last logged: 18:04:26

Error doing job: Task exception was never retrieved (task: None)
Traceback (most recent call last):
  File "/config/custom_components/bambu_lab/coordinator.py", line 273, in _handle_service_call_event
    result = self._service_call_unload_filament(data)
  File "/config/custom_components/bambu_lab/coordinator.py", line 639, in _service_call_unload_filament
    return self._service_call_load_unload_filament(False, data)
[...]

PR #1843, opened 2026-01-14
Update unload_filament to take in entity_id instead of device_id to f…

ZeldoKavira:
…ix regression from November

## Description

Fix to unload filament from when it was swapped to entity_id, but the definition was not updated

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Documentation

## Testing


AdrianGarside:
Hmm, this is not right. I need to review what's going on - the unload action applies to the whole device so this one needs to remain device_id.

AdrianGarside:
I've reactivated #1841 and will investigate it.

Issue #1848, opened 2026-01-19
[Bug] AMS 1 No Longer Showing Empty


TastePurple:
### Home Assistant Version

2026.1.2

### Bambu Lab Integration Version

2.2.19

### Describe the bug

AMS 1 used to show "Empty" when the spool was removed but now it retains the last filament and percentage. The Bambu iOS app shows the slot as "Empty", the Bambu Studio shows the slot as "Empty", but the HA integration did not get the update.

Bambu Studio
<img width="274" height="180" alt="Image" src="https://github.com/user-attachments/assets/9d80df9a-b7c3-4ca4-9da0-a6cd05418ea1" />

iOS
<img width="451" height="180" alt="Image" src="https://github.com/user-attachments/assets/5aa2680b-350b-4505-b58e-aaa6468d2c8e" />

Home Assistant
<img width="356" height="180" alt="Image" src="https://github.com/user-attachments/assets/de17d505-9dc6-48ee-842f-fb276ca1a017" />

### To Reproduce

1. Load a spool into the AMS
2. Confirm the slot is reported correctly in Bambu and HA interface
3. Remove the spool
4. Confirm the slot is empty in Bambu interface but retaining prior value in HA interface.

### Expected Behaviour

The HA interface should display "Empty" on the slots that are empty.

### What device are you using?

AMS

### Other Information

The AMS HT on the same setup is functioning properly with regard to displaying "Empty" when the spool has been removed.

AdrianGarside:
https://github.com/greghesp/ha-bambulab/releases/tag/v2.2.20-beta1

TastePurple:
That was incredibly fast! Thank you very much for that. I can confirm that fixed the issue.

gibz104:
Hi, thanks for the quick fix in v2.2.20-beta1! I have a question about the expected behavior.

I'm running v2.2.20-beta1 on an X1C in cloud mode. I noticed that:

1. When I physically remove a spool, the tray sensor does not update in real-time - it continues showing the previous filament info
2. When I manually reload the integration (Settings → Devices & Services → Bambu Lab → Reload), the tray correctly shows "Empty"

Is this expected behavior? I wasn't sure if real-time detection of spool removal was intended to work, or if the fix was specifically for ensuring correct state on reload/restart.

Steps to reproduce:
1. Have a spool loaded in a tray (shows correctly in HA)
2. Physically remove the spool from the AMS
3. Observe HA - tray still shows previous filament
[...]

PR #1855, opened 2026-01-25
Fix empty tray detection for real-time updates

gibz104:
## Description

The v2.2.20-beta1 fix for empty tray detection works on integration reload but not during real-time MQTT updates.                                                        
                                                                                                                                                                           
Root cause: The current check (len(data) == 1) and ('id' in data) fails because empty trays send 2 fields (id and state), not 1. Additionally, data.get('field', self.field) fallbacks retain stale filament data when trays become empty.

MQTT evidence - Empty tray message from printer:
{"id": "3", "state": 10}                                                                                                                                                 
                                                                                                                                                                         
This fix:                                                                                                                                                                
1. Detects empty by checking for absence of filament data: ('tray_type' not in data) and ('tray_info_idx' not in data)                                                   
2. Processes empty check FIRST before other field updates                                                                                                                
3. Explicitly resets all fields to empty defaults when tray is empty                                                                                                     
4. Only updates filament fields in the else branch when tray contains a spool

## Type of Change

<!-- Mark the appropriate option(s) with an 'x' -->

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### Link to Issue

https://github.com/greghesp/ha-bambulab/issues/1848

## Documentation

<!-- Mark the following checklist with an 'x' to confirm -->

## Testing

<!-- Describe the testing you have performed -->

 Manually tested on X1C in cloud mode:                                                                                                                                    
  - Physically removed spool → tray immediately shows "Empty" in HA (real-time)                                                                                            
  - Inserted spool → tray correctly populates filament info (real-time)                                                                                                    
  - Integration reload → correct state preserved                                                                                                                           
  - No regressions observed

## Additional Notes

<!-- Add any additional information that reviewers should know -->

The reason reload worked but real-time didn't: On reload, AMSTray objects are re-created with empty defaults in __init__, so stale data wasn't an issue. During real-time MQTT updates, existing objects retained previous values due to the data.get('field', self.field) fallback pattern.

### DIFF
--- a/custom_components/bambu_lab/pybambu/models.py
+++ b/custom_components/bambu_lab/pybambu/models.py
@@ -2696,24 +2696,11 @@ class AMSTray:
     def print_update(self, data) -> bool:
         old_data = f"{self.__dict__}"
 
-        self.idx = data.get('tray_info_idx', self.idx)
-        self.name = get_filament_name(self.idx, self._client.slicer_settings.custom_filaments)
-        self.type = data.get('tray_type', self.type)
-        if self.name == "unknown":
-            # Fallback to the type if the name is unknown
-            self.name = self.type
-        self.sub_brands = data.get('tray_sub_brands', self.sub_brands)
-        self.color = data.get('tray_color', self.color)
-        self.nozzle_temp_min = data.get('nozzle_temp_min', self.nozzle_temp_min)
-        self.nozzle_temp_max = data.get('nozzle_temp_max', self.nozzle_temp_max)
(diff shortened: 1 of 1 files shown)


AdrianGarside:
Unfortunately this won’t work. Lack of data doesn’t mean empty when the printer is sending the Mqtt payload only as deltas. That’s why detecting empty in a way that works for both delta payload and full payloads is hard.

PR #1861, opened 2026-01-27
Revise 3MF printing instructions and file path

lfpoulain:
Updated description for printing 3MF files and corrected file path.

## Description

1.Removed "cache/" from the filepath - The file is at the SD card root, not in a cache folder.
2.Changed the file extension from ".3mf" to ".gcode.3mf" - Bambu Studio actually exports files with the ".gcode.3mf" extension, both for local export and when sending to printer in LAN mode.
3.Added an explanatory note - Clarified that Bambu Studio outputs ".gcode.3mf" files in both export scenarios.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Documentation

- [x] I have updated the relevant documentation to reflect these changes

## Testing

### DIFF
--- a/docs/actions.mdx
+++ b/docs/actions.mdx
@@ -43,13 +43,12 @@ or not so before invoking this action, you should check the printer state to ens
 ## Print 3MF project file
 
 <Property name="action" type="bambu_lab.print_project_file" required>
-Prints 3MF file stored on SD card. 3MF file must be stored on SD card root and it must contain gcode (e.g., "Export sliced file" from slicer)
-
-    ```yaml
+Prints 3MF file stored on SD card. 3MF file must be stored on SD card root and it must contain gcode (e.g., "Export sliced file" from Bambu Studio). Note: Bambu Studio outputs files with the extension ".gcode.3mf" both for local export and when sending via LAN mode.
+```yaml
     action: bambu_lab.print_project_file
     data:
       device_id: a1b2c3d4e5f6g7h8i9j0
-      filepath: cache/filename.3mf
+      filepath: filename.gcode.3mf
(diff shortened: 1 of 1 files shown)


AdrianGarside:
The file can be in the root or in a cache directory (or another directory for that matter). The cache directory is, if anything, more common, so your revisions to the instructions are incorrect. Bambu Studio sends the file both as .3mf and .gcode.3mf depending on the printer model.

AdrianGarside:
I'm happy to improve the instructions but it is a lot more variable than what you are seeing.

lfpoulain:
Ok I'll try more configuration on my side and will commit again. There is still "must be in root folder" and "gcode.3mf" that should be changed for clarity 😃

AdrianGarside:
It doesn’t have to be in the root folder. The path has to match where it is. It doesn’t have to be .gcode.3mf. It has to match the actual file name which can be .3mf. It depends on the printer model and how you sent the print (print or send for Bambu Studio or Orca Slicer or makerworld), whether the printer is in lan mode or not, whether the printer has sd/usb connected (when optional). The example command is 100% correct for a P1S and print sent from Bambu Studio in non lan mode.

lfpoulain:
Sorry if I misunderstood something, but it clearly says “the 3MF file must be stored on the SD card root” in the example. The example uses /cache, and you’re telling me it could be anywhere. I think a better explanation, like the one you gave me, could be in the docs :)

AdrianGarside:
Ah yes, you are correct. I thought you had added that in this delta in my overly quick review. That does indeed need correcting.

PR #1865, opened 2026-01-30
Fix broken card registration with 2026.2

piitaya:
## Description

In 2026.2, the parameter has been renamed to `resource_mode` as it now only have effect on the resources as the default yaml dashboard configuration is gone.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Documentation

## Testing

I tested on my instance using 2026.2.0b1 version.


Uber1337NL:
Tested it. That indeed fixed the issue introduces in 2026.2. Thanks for the quick-fix Paul!

PR #1910, opened 2026-03-02
Use translation placeholder for tray

piitaya:
## Description

Consolidate duplicated tray translation entries (`tray_1`, `tray_2`, `tray_3`, `tray_4`) into a single `tray` entry using Home Assistant's `translation_placeholders` feature with a `{tray_number}` placeholder.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Other (please describe): Translation refactoring / deduplication

## Documentation

- [ ] I have updated the relevant documentation to reflect these changes
- [x] No documentation updates were necessary for this change

## Testing

- [ ] I have added/updated tests that prove my fix is effective or that my feature works
- [ ] All new and existing tests passed

## Additional Notes

Entity unique IDs are unchanged (still based on `key` which remains `tray_1`, `tray_2`, etc.), so this is a non-breaking change. Only the `translation_key` and translation files are affected.

PR #1930, opened 2026-03-16
Use translation placeholder for tray

piitaya:
## Description

Consolidate duplicated tray translation entries (tray_1, tray_2, tray_3, tray_4) into a single tray entry using Home Assistant's translation_placeholders feature with a {tray_number} placeholder.

It replaces https://github.com/greghesp/ha-bambulab/pull/1910. The slot addition will allow usage in the cards.

Needs https://github.com/greghesp/ha-bambulab-cards/pull/143 before merging to avoid breaking change.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Other (please describe): Translation refactoring / deduplication

## Documentation

## Testing


AdrianGarside:
What will need changing on the card side to be able to enumerate these successfully with this change?

The current logic is:

```
const ENTITYLIST: string[] = [
  "ams_temp",
  "temp",             // Node-RED only
  "custom_humidity",
  "humidity_index",
  "humidity_level",   // Node-RED only
  "tray_0",           // Node-RED only
  "tray_1",
  "tray_2",
  "tray_3",
  "tray_4",
];

export function getBambuDeviceEntities(
  hass,
  device_id,
  entities: string[]
): { [key: string]: Entity } {
  const result: { [key: string]: Entity } = {};
  // Loop through all hass entities, and find those that belong to the selected device
  for (let k in hass.entities) {
    const value = hass.entities[k];
    if (value.device_id === device_id) {
      for (const key of entities) {
        if (value.platform == "bambu_lab") {
          const divider = key.indexOf('|')
          if (divider != -1) {
            const type = key.slice(0, divider)
            const name = key.slice(divider + 1)
            if (value.entity_id.startsWith(type) && name == value.translation_key) {
              result[name] = value;
            }
          } else {
            if (key == value.translation_key) {
              result[key] = value;
            }
          }
        } else if (value.platform == "mqtt") {
          let regex;
          if (key.startsWith("^")) {
            regex = new RegExp(key);
          } else {
            regex = new RegExp(`.*${key}$`);
          }
          if (regex.test(value.entity_id)) {
            result[key] = value;
          }
        }
      }
    }
  }
  return result;
}
```

So given a list of translation keys (and optionally an entity time if necessary to disambiguate), it can synchronously find all the matching entities. And then the passed back list uses those same translation keys as identifiers for the entities.

piitaya:
I edited the logic in the cards repo. It is enough, right? https://github.com/greghesp/ha-bambulab-cards/pull/143

AdrianGarside:
Aha, I hadn't noticed the cards PR :)

PR #1953, opened 2026-03-31
Expose additional sensors, add more accessory modules for H2 series

rhooper:
## Description

Exposes additional sensor values, splits out the accessories for the H2 series into their own devices.  Looking for feedback before finalizing.

## Type of Change

Adds

- [ ] Bug fix
- [X] New feature
- [?] Breaking change
- [ ] Documentation update

### Link to Issue

n/a

## Documentation

## Testing

- [X] I have added/updated tests that prove my fix is effective or that my feature works

## Additional Notes

Claude was used here.

### DIFF
--- a/custom_components/bambu_lab/binary_sensor.py
+++ b/custom_components/bambu_lab/binary_sensor.py
@@ -12,12 +12,17 @@ from .definitions import (
     AMS_BINARY_SENSORS,
     PRINTER_BINARY_SENSORS,
     VIRTUAL_TRAY_BINARY_SENSORS,
+    FIRE_EXTINGUISHER_BINARY_SENSORS,
+    LASER_BINARY_SENSORS,
+    ROTARY_BINARY_SENSORS,
     BambuLabAMSBinarySensorEntityDescription,
     BambuLabBinarySensorEntityDescription,
 )
 from .models import (
     AMSEntity,
     BambuLabEntity,
+    FireExtinguisherEntity,
+    ToolModuleEntity,
     VirtualTrayEntity,
 )
 from .pybambu.const import Features
@@ -34,7 +39,22 @@ async def async_setup_entry(
     if not coordinator.get_model().has_full_printer_data:
         return
 
-    for sensor in PRINTER_BINARY_SENSORS:    
+    if coordinator.get_model().supports_feature(Features.FIRE_EXTINGUISHER):
+        for sensor in FIRE_EXTINGUISHER_BINARY_SENSORS:
+            if sensor.exists_fn(coordinator):
+                async_add_entities([BambuLabFireExtinguisherBinarySensor(coordinator, sensor)])
+
+    for module in coordinator.get_model().extruder_tool.get_modules_by_type("laser"):
+        for sensor in LASER_BINARY_SENSORS:
+            if sensor.exists_fn(coordinator):
+                async_add_entities([BambuLabToolModuleBinarySensor(coordinator, sensor, module.serial)])
+
+    for module in coordinator.get_model().extruder_tool.get_modules_by_type("rotary"):
+        for sensor in ROTARY_BINARY_SENSORS:
+            if sensor.exists_fn(coordinator):
+                async_add_entities([BambuLabToolModuleBinarySensor(coordinator, sensor, module.serial)])
+
+    for sensor in PRINTER_BINARY_SENSORS:
         if sensor.exists_fn(coordinator):
             async_add_entities([BambuLabBinarySensor(coordinator, sensor)])
 
@@ -72,6 +92,11 @@ class BambuLabBinarySensor(BambuLabEntity, BinarySensorEntity):
     def is_on(self) -> bool:
         """Return if binary sensor is on."""
         return self.entity_description.is_on_fn(self)
+
+    @property
+    def available(self) -> bool:
+        """Return if entity is available."""
+        return self.entity_description.available_fn(self)
 
     @property
     def extra_state_attributes(self) -> dict:
@@ -114,3 +139,35 @@ class BambuLabExternalSpoolBinarySensor(VirtualTrayEntity, BambuLabBinarySensor)
         self.suffix = suffix
         self.entity_description = description
         self._attr_unique_id = f"{printer.device_type}_{printer.serial}_ExternalSpool{suffix}_{description.key}"
+
+
+class BambuLabFireExtinguisherBinarySensor(FireExtinguisherEntity, BambuLabBinarySensor):
+    """Representation of a BambuLab Fire Extinguisher binary sensor."""
+
+    def __init__(
+            self,
+            coordinator: BambuDataUpdateCoordinator,
+            description: BambuLabBinarySensorEntityDescription,
+    ) -> None:
+        """Initialize the sensor."""
+        super().__init__(coordinator=coordinator, description=description)
+        printer = coordinator.get_model().info
+        self.entity_description = description
+        self._attr_unique_id = f"{printer.device_type}_{printer.serial}_FireExtinguisher_{description.key}"
+
+
+class BambuLabToolModuleBinarySensor(ToolModuleEntity, BambuLabBinarySensor):
+    """Representation of a BambuLab Tool Module binary sensor."""
+
+    def __init__(
+            self,
+            coordinator: BambuDataUpdateCoordinator,
+            description: BambuLabBinarySensorEntityDescription,
+            module_serial: str,
+    ) -> None:
+        """Initialize the sensor."""
+        super().__init__(coordinator=coordinator, description=description)
+        self.module_serial = module_serial
+        printer = coordinator.get_model().info
+        self.entity_description = description
+        self._attr_unique_id = f"{printer.device_type}_{module_serial}_{description.key}"
--- a/custom_components/bambu_lab/definitions.py
+++ b/custom_components/bambu_lab/definitions.py
@@ -102,6 +102,18 @@ class BambuLabHotendRackSensorEntityDescription(
 
 
 @dataclass
+class BambuLabFireExtinguisherSensorEntityDescription(
+    SensorEntityDescription, BambuLabSensorEntityDescriptionMixin
+):
+    """Sensor entity description for Bambu Lab Fire Extinguisher."""
+
+    available_fn: Callable[..., bool] = lambda _: True
+    exists_fn: Callable[..., bool] = lambda _: True
+    extra_attributes: Callable[..., dict] = lambda _: {}
+    icon_fn: Callable[..., str] = lambda _: None
+
+
+@dataclass
 class BambuLabBinarySensorEntityDescriptionMixIn:
     """Mixin for required keys."""
     is_on_fn: Callable[..., bool]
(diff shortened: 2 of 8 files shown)


custom_components/bambu_lab/pybambu/models.py
```diff
@@ -3550,13 +3535,160 @@ def update(self):
...
+            fire_ext = data["device"]["fire_ext"]
+            for attr, key in [
+                ("_state", "state"),
+                ("_connect_flag", "connect_flag"),
+                ("_remain_time", "remain_time"),
+                ("_rm_init", "rm_init"),
+                ("_bot_err", "bot_err"),
+                ("_cd", "cd"),
+                ("_cd_init", "cd_init"),
+                ("_drill_flag", "drill_flag"),
```
AdrianGarside, line 3599: Are all of these valuable to expose? It looks like you aren't using all of them?

custom_components/bambu_lab/pybambu/models.py
```diff
@@ -3575,9 +3707,26 @@ def print_update (self, data):
...
+                self.state = "cooling_fan"
             elif mount == 1 and tool_type:
-                self.state = None
-        
+                self.state = "unknown"
+
+            # Parse additional fields
+            if "th_temp" in ext_tool:
+                self.th_temp = ext_tool["th_temp"]
+            if "calib" in ext_tool:
```
AdrianGarside: This can be made more compact to handle the delta mqtt payloads as:
self.calib = ext.tool.get("calib", self.calib)

custom_components/bambu_lab/binary_sensor.py
```diff
@@ -34,7 +39,22 @@ async def async_setup_entry(
...
+        for sensor in FIRE_EXTINGUISHER_BINARY_SENSORS:
+            if sensor.exists_fn(coordinator):
+                async_add_entities([BambuLabFireExtinguisherBinarySensor(coordinator, sensor)])
+
+    for module in coordinator.get_model().extruder_tool.get_modules_by_type("laser"):
+        for sensor in LASER_BINARY_SENSORS:
+            if sensor.exists_fn(coordinator):
+                async_add_entities([BambuLabToolModuleBinarySensor(coordinator, sensor, module.serial)])
+
+    for module in coordinator.get_model().extruder_tool.get_modules_by_type("rotary"):
```
AdrianGarside, line 52: Are we certain this data has arrived and been processed by the time the entity creation runs. I think we are given the way I reworked initialization. But this should be double checked or we'll be missing sensors, potentially only some of the time.

custom_components/bambu_lab/definitions.py
```diff
@@ -919,3 +900,85 @@ def _hotend_sensor(slot_id: int, display_id: int) -> BambuLabHotendRackSensorEnt
...
+# Fire Extinguisher sensors (on Fire Extinguisher device)
+FIRE_EXTINGUISHER_SENSORS: tuple[BambuLabFireExtinguisherSensorEntityDescription, ...] = (
+    BambuLabFireExtinguisherSensorEntityDescription(
+        key="fire_ext_status",
+        translation_key="fire_ext_status",
+        icon="mdi:fire-extinguisher",
+        device_class=SensorDeviceClass.ENUM,
+        options=["standby", "active"],
+        value_fn=lambda self: self.coordinator.get_model().fire_extinguisher.status,
+        available_fn=lambda self: self.coordinator.get_model().fire_extinguisher.connected,
```
AdrianGarside, line 947: In general I'm trying to move away from making entities unavailable (due to people that complain about all unavailable entities). If not connected, is that a permanent state or can it change live? If it's more permanent - this should be driving the exists check. But none of these sensors have an exists_fn even though you are trying to call that - which means it's always going to be returning true.

custom_components/bambu_lab/binary_sensor.py
rhooper: Seemed to be okay... What specific concern can i look into to see if this is badly structured.

custom_components/bambu_lab/definitions.py
rhooper: The fire extinguisher, rotary tool, external pump, cutting tool, and laser tool are all hot connectable.  Is the correct fix to implement exists_fn?

custom_components/bambu_lab/pybambu/models.py
rhooper: In my updated (and rebased) push, they're now exposed so I can monitor for their values.

custom_components/bambu_lab/pybambu/models.py
rhooper: I rewrote it to use this style, then claude pointed out that I'd be clobbering values on partial mqtt updates causing oscillations between None and being set.

custom_components/bambu_lab/definitions.py
AdrianGarside: No, if they're hot connectable then unavailable is probably the best option. But you do need to implement exists_fn because most of the bambu models do not have this and never will. exists needs to return false for those models.

custom_components/bambu_lab/sensor.py
```diff
@@ -56,6 +59,16 @@ async def async_setup_entry(
             if sensor.exists_fn(coordinator):
                 async_add_entities([BambuLabHotendSensor(coordinator, sensor)])
 
+    if coordinator.get_model().supports_feature(Features.FIRE_EXTINGUISHER):
+        for sensor in FIRE_EXTINGUISHER_SENSORS:
+            if sensor.exists_fn(coordinator):
+                async_add_entities([BambuLabFireExtinguisherSensor(coordinator, sensor)])
+
+    for module in coordinator.get_model().extruder_tool.get_modules_by_type("laser"):
```
AdrianGarside, line 67: Needs a supports_feature check to limit this to models that have this

custom_components/bambu_lab/definitions.py
AdrianGarside: You have supports_features check at device creation - move that into the exists check instead.

custom_components/bambu_lab/sensor.py
AdrianGarside: Or preferably implement the exists_fn method and add the feature check there.

AdrianGarside (changes requested):

custom_components/bambu_lab/definitions.py
pkolbus: On this topic -- are the devices only created when they are first seen as connected? A laser or fire extinguisher device doesn't make sense for H2 owners that don't have the laser edition/upgrade. 

If you'd like me to test (I have an H2D and bought the cutter on sale last year), let me know when you get an update made.

pkolbus:
> [] I have updated the relevant documentation to reflect these changes

At least the entities doc would benefit from an update; maybe more as the list of devices is growing.

custom_components/bambu_lab/definitions.py
AdrianGarside: Yes, we can probably do better than that - I re-initialize sensors in certain scenarios already - such as when the user toggles the camera on/off so that the entity goes away when it's off. We'll need an event that the pybambu layer raises that would trigger the coordinator to refresh the entities. Then that event can be raised each time a hardware change occurs like this.

AdrianGarside: Would be more work to extend that to an entire device though - I have logic to detect and delete stale AMS's. It should be possible in principle though. Not sure whether entire devices coming/going would be the best experience though.

pkolbus: This is getting off topic from the review. Maybe we should take the discussion elsewhere.

For background, I've been using Home Assistant for years; my trade is software engineering and cybersecurity. My instance is on the larger side -- about 160 automations,  80 user-facing integrations (21 custom), 8000+ entitles, 4600+ enabled. At this scale, the more consistently integrations behave the better; the [Integration Quality Scale](https://developers.home-assistant.io/docs/core/integration-quality-scale/) is the HA developers' vision of what a good integration is and what the built-in integrations are using to advance the experience. My comments are rooted in that scale.

"Unavailable" is well-specified in [the quality scale](https://developers.home-assistant.io/docs/core/integration-quality-scale/rules/entity-unavailable) and likely other documentation/community posts, it is a legitimate state and means that the entity is known but data cannot be fetched.  (Which lines up with a tool's entities when that tool has been seen but is not connected.) This state has a few rough edges, but there is enough in HA to manage it. Updating my templates and automations to consider 'unavailable' has made my system more reliable (in terms of not doing weird things.)

The ideal time to create devices is [when they are connected](https://developers.home-assistant.io/docs/core/integration-quality-scale/rules/dynamic-devices). This is a Gold-level rule and commissioning new devices is relatively uncommon; if a new device doesn't show up immediately I'm OK with restarting Home Assistant or the integration to make it work.

The right time to remove devices/entities is also [specified](https://developers.home-assistant.io/docs/core/integration-quality-scale/rules/stale-devices): when the integration *knows* the device is no longer present. When there's a hub or account this is clear as the hub/account can be interrogated. It's less clear for a Bambu Lab printer when there's only a LAN connection. As this is another Gold-level rule and decommissioning is uncommon, it's OK if I have to manually remove devices that I've stopped using. I'd rather be intentional about that than have entities I depend on disappear.

Taking a couple of cases from ha-bambulab:
* An entity that's always connected to the printer, like the camera, is unlikely to be decommissioned. The user might have turned off LiveView at the printer, but after a power cycle, it's *much* more likely the camera has not yet initialized (or failed). Home Assistant keeps the camera in the entity registry, so delayed creation is not fatal, but in that case the entity is not included in (integration_entities())[https://www.home-assistant.io/template-functions/integration_entities/].
* For pluggable accessories that I own like AMSes,, cutter, etc. I'd expect the devices to remain in Home Assistant indefinitely. This avoids problems with running scripts/automations, and enables working on automations without the device connected.  Even if the accessory is not seen for months or many runtime hours, it could be that I only use that tool a few times a year. Offering a repair might be appropriate though, ss it leaves the decision to the user.

(P.S. Even further off topic, I have a repair automation that reloads bambu_lab if the camera remains unavailable, so I don't know if the camera delay is the integration being dynamic or just my reload. Either way it's working well enough for me, now.)

custom_components/bambu_lab/pybambu/models.py
AdrianGarside: self.calib = ext.tool.get("calib", self.calib)
This means get("calib") and if not present, set self.calib = self.calib (no change)
So I don't know what Claude was complaining about.

custom_components/bambu_lab/definitions.py
AdrianGarside: The AMS stale detection is because it's legal and not that uncommon for people to reconfigure AMSs between their printers. If I don't remove them from the printers they are no longer connected to, things end up being very broken for that user. IIRC I only remove an AMS when it's been seen connected to the current instance and I can see orphaned connections to other instances of the integration (but I'll check - I may be misremembering). Which I think should handle both scenarios - an AMS going offline for a while shouldn't disappear but moving an AMS should move.

Are the tools uniquely identified such that they can also move between printers?

AdrianGarside: Deferring creation of the devices until they are connected is definitely the best choice as I suspect the vast majority of owners will never get them.

PR #1976, opened 2026-04-22
Improve wiki links

pkolbus:
## Description

HMS troubleshooting pages on the Bambu Lab Wiki are not series-specific, but vary in breadth of applicability. In some cases (e.g. HMS_0300-0100-0001-0003), multiple series share the same URL, and the URL has a representative or first model. Other HMS codes are model-specific, e.g. h2s and h2d have different codes. Additionally, a few troubleshooting pages have hyphens instead of underscores in the URL.

https://wiki.bambulab.com/en/hms/home has links to troubleshooting pages, with the text identifying the model names that the link applies to. This PR adds a scraper to `update_error_text` to generate a new wiki_links.json.gz using the same compression format as the HMS error texts, and updates pybambu.utils to use the wiki_links data. Existing tests are updated and now pass.

In cases where there is no known Wiki page for an HMS code, the HMS home page https://wiki.bambulab.com/en/hms/home is used instead; this is more useful than a 404 error.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Documentation

## Testing


pkolbus:
There is one additional aspect that might improve the URL quality -- the HMS home page https://wiki.bambulab.com/en/hms/home lists synonyms for HMS codes. For example, the X1 page for 0300_9100_0001_0003 also helps for 0300_9200_0001_0003. This isn't a regression from previous behavior but something I could work on either as another patch here or another PR.

PR #2001, opened 2026-05-09
fan_percentage: use math.ceil to stop fan-speed entities flipping between 10% buckets

jesung:
## Description

`pybambu/utils.py:39` `fan_percentage()` bucketed raw 0–15 PWM values
to 10% steps using `round()`. The printer reports an instantaneous
raw value that naturally alternates between adjacent integers when
serving a target whose exact 0–15 representation isn't an integer
(e.g. raw 2 and raw 3 both serve a duty cycle around 20%). With
`round()`, HA fan-speed entities flipped between adjacent 10% buckets
on every MQTT push:

```
raw 2 -> (2/15)*100 = 13.33 -> round(1.33)*10 = 10
raw 3 -> (3/15)*100 = 20.00 -> round(2.00)*10 = 20
```

Swap to `math.ceil` and both raw values land in the same bucket
(20%), eliminating the oscillation.

`definitions.py:49` already has a (currently unused) `fan_to_percent`
helper that uses `math.ceil` for the same conversion, so this matches
an existing convention in the codebase.

### Live evidence

Captured from an X2D with the exhaust fan set to 20% via the printer
touchscreen. `big_fan2_speed` (the raw value backing
`chamber_fan_speed` for X2D-style printers) across 13 consecutive
print pushes:

```
 #  big_fan2  airduct[id=48].state
 1     '2'         20
 2     '2'         20
 3     '3'         20
 4     '2'         20
 5     '3'         20
 ...   ...         ...
13     '3'         20
```

Pre-fix this stream flipped `sensor.x2d_…_chamber_fan_speed` between
10% and 20% on alternating messages. Post-fix, polled every 2 s for
16 s with the printer continuing to oscillate the raw value:

```
sensor.x2d_…_chamber_fan_speed = 20  (last_updated unchanged)
```

Diagnostic dump confirms the underlying parser still sees the raw
oscillation while the percentage stays stable:

```
raw=_chamber_fan_speed='2'   pct=_chamber_fan_speed_percentage=20
raw=_chamber_fan_speed='3'   pct=_chamber_fan_speed_percentage=20
raw=_chamber_fan_speed='3'   pct=_chamber_fan_speed_percentage=20
```

### Affected entities

All entities using `fan_percentage()` — i.e. all `*_fan_speed`
sensors via `Fans.print_update`:

- `aux_fan_speed`
- `chamber_fan_speed`
- `cooling_fan_speed`
- `heatbreak_fan_speed`

Not X2D-specific. Affects every model that reads fans through the
legacy `big_fan*_speed` / `cooling_fan_speed` / `heatbreak_fan_speed`
top-level fields.

### Trade-off

`math.ceil` causes the reported value to be `>=` the actual duty
cycle (e.g. raw 1 = 6.67% reported as 10%). `round()` could go either
way. Since the user-facing setpoint moves in 10% steps and the
printer's PWM resolution is coarser than the display, both choices
involve some lossiness; `ceil` is the choice that doesn't *flip* on
oscillation and never *under*-reports actual fan output.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### Link to Issue

No matching open issue found.

## Documentation

## Testing

Four new regression tests added in `pybambu/tests/test_utils.py`:

- `test_zero_speed_returns_zero` — boundary invariant
- `test_full_speed_returns_one_hundred` — boundary invariant
- `test_oscillation_between_raw_2_and_3_pins_to_same_bucket` — the
  core regression: fixed code yields same bucket for raw 2 and 3;
  buggy code yields 10 and 20 with a clear diagnostic message.
- `test_low_speeds_round_up_to_nearest_ten` — verifies `ceil`
  semantics across raw 1–5.

Verified red→green: ran against the deployed buggy `utils.py`
(2 of 4 fail with the expected mismatch, the 2 boundary tests pass
under both `round` and `ceil`); against the patched version 4/4 pass.

Full `pybambu/tests/` suite: **36/36 pass** against the patched code
(26 in `test_models.py` + 6 in `test_error_lookup.py` + 4 new in
`test_utils.py`).

End-to-end verification: patched `utils.py` deployed onto a running
HAOS instance and HA Core restarted (Python module cache means
`reload_config_entry` alone doesn't pick up the change). Polled
`sensor.x2d_…_chamber_fan_speed` every 2 s for 16 s while the
printer continued to alternate `big_fan2_speed` between '2' and '3' —
HA value held at 20% with no state-change updates emitted. Pre-patch
behaviour on the same hardware was the alternating 10%↔20% flip.

## Additional Notes

The dead `fan_to_percent` in `definitions.py:49` (no callsites in the
codebase) already used `math.ceil` for what looks like the same
conversion, so a parallel maintainer view of "what should the
rounding be" already arrived at `ceil`. Happy to delete that dead
function in this PR or keep it for a tidy-up later — your call.

### DIFF
--- a/custom_components/bambu_lab/pybambu/tests/test_utils.py
+++ b/custom_components/bambu_lab/pybambu/tests/test_utils.py
@@ -1,3 +1,4 @@
+import unittest
 from unittest.mock import MagicMock
 import json
 import os
@@ -7,7 +8,49 @@ from typing import Dict, Any, Callable, Optional
 from ..const import (
     LOGGER,
 )
-from ..utils import safe_json_loads
+from ..utils import safe_json_loads, fan_percentage
+
+
+class TestFanPercentage(unittest.TestCase):
+    """Regression tests for the fan_percentage round->ceil fix.
+
+    The printer reports an instantaneous raw 0-15 PWM value that
+    oscillates between adjacent integers when serving a target whose
+    exact 0-15 representation isn't an integer. round() picks
+    different buckets either side of the midpoint, so HA fan-speed
+    entities flip between adjacent 10% values on every push.
+    """
+
+    def test_zero_speed_returns_zero(self):
+        self.assertEqual(fan_percentage(0), 0)
+        self.assertEqual(fan_percentage("0"), 0)
+        self.assertEqual(fan_percentage(None), 0)
+
+    def test_full_speed_returns_one_hundred(self):
+        self.assertEqual(fan_percentage(15), 100)
+        self.assertEqual(fan_percentage("15"), 100)
+
+    def test_oscillation_between_raw_2_and_3_pins_to_same_bucket(self):
+        # Live capture: with user setting at 20% the printer alternates
+        # big_fan2_speed between '2' and '3' across consecutive print
+        # pushes. Both must round to the same 10% bucket so HA doesn't
+        # flip on every message.
+        speed_for_raw_2 = fan_percentage(2)
+        speed_for_raw_3 = fan_percentage(3)
+        self.assertEqual(
+            speed_for_raw_2, speed_for_raw_3,
+            f"raw 2 -> {speed_for_raw_2} but raw 3 -> {speed_for_raw_3}; "
+            "values disagree => HA will flip on every MQTT push"
+        )
+
+    def test_low_speeds_round_up_to_nearest_ten(self):
+        # raw 1 -> 6.67%, raw 2 -> 13.33%, raw 4 -> 26.67%
+        # All should round up so the reported value is never lower
+        # than the actual fan output.
+        self.assertEqual(fan_percentage(1), 10)
+        self.assertEqual(fan_percentage(2), 20)
+        self.assertEqual(fan_percentage(4), 30)
+        self.assertEqual(fan_percentage(5), 40)
 
 class MqttMessageInfo:
     """Mock MQTT message info object."""
--- a/custom_components/bambu_lab/pybambu/utils.py
+++ b/custom_components/bambu_lab/pybambu/utils.py
@@ -37,11 +37,20 @@ def search(lst, predicate, default={}):
 
 
 def fan_percentage(speed):
-    """Converts a fan speed to percentage"""
+    """Converts a fan speed to percentage.
+
+    Bambu fans report a raw 0-15 PWM-step value. The user-facing
+    setting moves in 10% increments, but the printer's reported
+    instantaneous value naturally oscillates between adjacent raw
+    integers when delivering an effective duty cycle (e.g. raw 2 and
+    raw 3 both serve a target around 20%). math.ceil keeps the
+    reported value pinned to the upper bucket so HA fan entities
+    don't flip 10% <-> 20% on every push.
+    """
     if not speed:
         return 0
     percentage = (int(speed) / 15) * 100
-    return round(percentage / 10) * 10
+    return math.ceil(percentage / 10) * 10
 
 
 def fan_percentage_to_gcode(fan: FansEnum, percentage: int):


jesung:
**Edit (2026-05-09):** I empirically swept all 10 user-set targets on a live X2D in heating mode and the picture is much cleaner than my original theoretical analysis suggested. Replacing the comment body with the actual data.

## Empirical sweep on a live X2D

For each user-set exhaust-fan target, captured ~10–12 consecutive print messages and observed the raw `big_fan2_speed` distribution + computed bucket result under both `round` (current) and `math.ceil` (this PR):

| User target | Raw distribution | `round` (current) | `ceil` (this PR) |
|---|---|---|---|
| 10% | raw `1` stable | 10% ✓ stable | 10% ✓ stable |
| 20% | raw `2` / `3` (oscillates) | **flips 10↔20** | 20% ✓ stable |
| 30% | raw `4` stable | 30% ✓ stable | 30% ✓ stable |
| 40% | raw `5` / `6` (oscillates) | **flips 30↔40** | 40% ✓ stable |
| 50% | raw `7` stable | 50% ✓ stable | 50% ✓ stable |
| 60% | raw `8` stable | **50% wrong** | 60% ✓ stable |
| 70% | raw `10` stable | 70% ✓ stable | 70% ✓ stable |
| 80% | raw `11` / `12` (oscillates) | **flips 70↔80** | 80% ✓ stable |
| 90% | raw `13` stable | 90% ✓ stable | 90% ✓ stable |
| 100% | raw `14` / `15` (oscillates) | **flips 90↔100** | 100% ✓ stable |

**On the X2D, `ceil` is correct and stable at all 10 targets.** `round` is broken at 5 of 10:

- Flips at 20/40/80/100 (oscillating raw values straddle a `round` bucket boundary)
- Mis-reports 60% as 50% (raw 8 = 53.3% rounds down to 50)

## Why my original "trade-off" framing was wrong

I had assumed the firmware would oscillate between adjacent integers at every "half-integer target" (10/30/50/70/90, where target × 15 / 100 isn't a clean integer). Empirically that's not what the X2D firmware does — it picks a stable raw value at every half-integer target (10/30/50/70/90) and oscillates instead at *some* of the "clean integer" targets (20/40/80/100). The oscillation pattern doesn't match a simple "averaging around the exact target" model.

## Updated recommendation

This PR is an unambiguous improvement on the X2D firmware that motivated it. The remaining concern is whether **other Bambu printer models** (A1 / P1S / X1C / X1E / H2x / P2S) oscillate on a different pattern that would make `round` better than `ceil` for them. I don't have empirical data on those — would appreciate maintainer or community confirmation, especially for any model where `round`'s 30↔30 stability at raw 4↔5 is observed in practice.

Two paths forward:

1. **Land this PR.** Strict win on X2D; theoretically *might* be a sideways trade on other models we haven't tested, but the worst case there is still no regression in flip count, just a shift in which targets flip. Defaults from `definitions.py:49`'s already-existing `fan_to_percent` (which uses `math.ceil`) suggest a maintainer-side preference for `ceil` semantics anyway.

2. **Replace with hysteresis** — only update `_*_fan_speed_percentage` when the new raw value differs from the cached previous-raw by ≥2. Universal across all printer models. Diff is somewhat larger because each fan needs a `_last_raw` field. Drafted as a follow-up if you'd prefer that route.

Either way the X2D-specific airduct sourcing PR (separate, planned) supersedes this question for X2D users — it reads `device.airduct.parts[].state` which is already in 10% steps and never oscillates.

I'm fine with whichever path you'd like to take. Apologies for the noise on the original framing.

AdrianGarside:
I remember having to tweak this to match the reported values from Bambu Studio/Handy. But then I think either they changed the firmware or each model behaves slightly differently. So, there's a chance this will fix the X2D but regress the A1/P1 models. But I prefer stable values over ones that oscillated and spam the HA entity history database.

Issue #1705, opened 2025-11-21
H2C: support needs to be implemented


VIDGuide:
So the H2C is detected and brings in a lot of the standard sensors. The "printer" card however cannot pick it as an entity, and the camera feed does not show up as an entity at all.

AdrianGarside:
Please provide a diagnostics log.

AdrianGarside:
Initial support changes:
https://github.com/greghesp/ha-bambulab/releases/tag/v2.2.13-beta7

VIDGuide:
Love your work mate! Just updated to the beta, and that's perfect, machine & camera working :)

ThomasLoeffel:
Hallo Zusammen,

ich kann den Drucker H2C über die Bambu-Cloud auswählen bekomme aber dann eine Fehlermeldung.
Leider habe ich keine Ahnung was ich tun soll. Ich kann Bambu Studio, oder die Iphone-App normal benutzen und dort funktionieren die Zugangsdaten wie er wartet.

Danke für Eure Hilfe.
Thomas

![Image](https://github.com/user-attachments/assets/4bcfc3cb-4bff-4b60-8785-7324725f0130)
![Image](https://github.com/user-attachments/assets/21716ee9-9ad9-40e4-9be8-0b141e5118ab)

AdrianGarside:
> Love your work mate! Just updated to the beta, and that's perfect, machine & camera working :)

Can I get a diagnostic log so I can create a mock to test against please.

ThomasLoeffel:
Was soll ich genau machen? Gerne sende ich Dir dann die Daten? Kannst Du mir eine Step by Step Anleitung geben?
Sorry und Danke, Thomas

AdrianGarside:
> Was soll ich genau machen? Gerne sende ich Dir dann die Daten? Kannst Du mir eine Step by Step Anleitung geben? Sorry und Danke, Thomas

Under the advanced section there is an option to turn off SSL validation. This should unblock you. However if you can, I would like to get the information about the different SSL certificate being used on this newest model. I'll look up the command I need running to get the SSL certificate details off the printer.

ThomasLoeffel:
Nur als Ergänzung, ich bekomme die Integration nicht installiert. Somit kann ich unter Integrationen nichts bei den drei Punkten (...) auswählen.

ThomasLoeffel:
[...]

PR #2002, opened 2026-05-12
refactor: use printer.cer bundle

j4k0xb:
## Description

https://github.com/bambulab/BambuStudio/blob/3e96c7e073fdc1470560c6874ba8ee64faa54e88/resources/cert/printer.cer was updated a while ago to include both the original `BBL CA` (for A1, P1, X1) and multiple `BBL CA2` (for newer ones) roots .

Previously this repo embedded the `BBL CA` and each series-specific intermediate cert. Using BambuStudio's bundle as the source of truth instead should make this more future-proof and avoid manual updates like #1611, #1707, etc. whenever a new printer series is released (A2 soon?).

Hierarchy:
- Root `BBL CA`:
  - Leaf `Device cert`: for each serial number
- Root `BBL CA2 RSA/ECC`
  - Intermediate `Series-specific cert`: H2, P2, X2, probably all newer ones in the coming years
    - Leaf `Device cert`: for each serial number

This PR remains full backwards-compatibility and [load_verify_locations](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_verify_locations) works with bundles.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [X] Other (please describe): Refactor/Future-proofing

## Documentation

## Testing

## Additional Notes

```sh
openssl verify -CAfile /opt/bambustudio-bin/resources/cert/printer.cer custom_components/bambu_lab/pybambu/certs/*
custom_components/bambu_lab/pybambu/certs/bambu.cert: OK
custom_components/bambu_lab/pybambu/certs/bambu_h2c_251122.cert: OK
custom_components/bambu_lab/pybambu/certs/bambu_p2s_250626.cert: OK
custom_components/bambu_lab/pybambu/certs/bambu_x2c_260425.cert: OK
```

also confirmed working against user's H2C chain from https://github.com/greghesp/ha-bambulab/issues/1705#issuecomment-3567083134

Technically all the other intermediate files like [certs/bambu_h2c_251122.cert](https://github.com/greghesp/ha-bambulab/tree/15ed3408df9e566340e32f323277dbd0b14d92c8/custom_components/bambu_lab/pybambu/certs) can be deleted now but I chose not to in case old printer firmware ever had a bug that shipped an incomplete TLS chain (for example only presenting the device leaf cert instead of series-specific cert -> device cert)... From the existing certs ppl posted this doesn't seem to be the case, anyone knows if this happened?


AdrianGarside:
You need to document the process for updating this combo cert - ideally with a script to do so.

j4k0xb:
Done. Added a script similar to https://github.com/greghesp/ha-bambulab/blob/f5193ed51fd3cb3f7d97f78666a22ae5b5714309/scripts/fetch_filaments_github.py

PR #2014, opened 2026-05-19
fix: prevent print-job entities from showing as unavailable when idle

BenHornerTech:
## Description

Fixes several entities becoming `unavailable` in Home Assistant when the printer is idle or a print finishes. Entities now remain available at all times, with safety guards added to prevent commands being sent when the printer isn't in the correct state.

**Buttons (Pause / Resume / Stop) - `button.py`:**
- Remove `available()` overrides so buttons inherit `CoordinatorEntity.available` (still correctly returns `False` when MQTT/connectivity is lost)
- Add state guards inside `async_press()` - commands are only published when the printer is in the correct gcode state; ignored presses log a warning

**Printing Speed Select - `select.py`:**
- Remove `available()` override that restricted the select to `gcode_state == 'RUNNING'`
- Add state guard inside `async_select_option()` checking for `RUNNING` or `PAUSE` before applying speed changes

**Print-job Sensors - `definitions.py`:**
- `start_time`: Set `available_fn` to always return `True`; existing `_start_time_value_fn()` already safely returns `None`
- `end_time`: Set `available_fn` to always return `True`; add null guard to `value_fn` to prevent `dt_util.as_utc(None)` raising an `AttributeError`
- `gcode_file` / `gcode_file_downloaded`: Set `available_fn` to always return `True`; return `None` instead of empty string when no file is set

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Documentation

## Testing

## Additional Notes

Claude was used to assist with these fixes and I have tested on my P1S with AMS.  I verified the entities remain available when idle, show `unavailable` only when the printer is genuinely offline, and buttons/speed select still function correctly during a print.

### DIFF
--- a/custom_components/bambu_lab/button.py
+++ b/custom_components/bambu_lab/button.py
@@ -138,7 +138,7 @@ class BambuLabResumeButton(BambuLabButton):
         return False
 
     async def async_press(self) -> None:
-        """ Pause the Print on button press"""
+        """ Resume the Print on button press"""
         self.coordinator.client.publish(RESUME)
 
 
--- a/custom_components/bambu_lab/definitions.py
+++ b/custom_components/bambu_lab/definitions.py
@@ -477,7 +477,7 @@ PRINTER_SENSORS: tuple[BambuLabSensorEntityDescription, ...] = (
         translation_key="start_time",
         icon="mdi:clock",
         device_class=SensorDeviceClass.TIMESTAMP,
-        available_fn=lambda self: self._start_time_avail_fn(),
+        available_fn=lambda self: True,
         value_fn=lambda self: self._start_time_value_fn(),
         is_restoring=True,
     ),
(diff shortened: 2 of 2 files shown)


custom_components/bambu_lab/button.py
```diff
@@ -113,50 +113,38 @@ class BambuLabPauseButton(BambuLabButton):
 
     entity_description = PAUSE_BUTTON_DESCRIPTION
 
-    @property
-    def available(self) -> bool:
```
AdrianGarside, line 117: This is very deliberate so that the cards can correctly show disabled buttons when the buttons cannot perform an action. Please remove this change.

AdrianGarside (changes requested):

custom_components/bambu_lab/button.py
BenHornerTech: Done.  I also corrected what I think is a copy-paste error in the ResumeButton docstring, as it previously said 'Pause' but the method publishes RESUME.

custom_components/bambu_lab/select.py
```diff
@@ -48,19 +48,17 @@ def __init__(self, coordinator: BambuDataUpdateCoordinator) -> None:
             speed for i, speed in SPEED_PROFILE.items()
         ]
 
-    @property
-    def available(self) -> bool:
-        """Return True if entity is available."""
-        return self.coordinator.get_model().print_job.gcode_state == 'RUNNING'
```
AdrianGarside, line 54: Ah, same thing here. The select needs to be unavailable for the cards to represent that state correctly.

AdrianGarside (changes requested):

custom_components/bambu_lab/select.py
BenHornerTech: Done, starting to feel like Home Assistant needs some way of representing non-actionable entities in a better way than just unavailable :-)

BenHornerTech: Actually looks like there is a proposal here that may eventually help: https://github.com/home-assistant/architecture/discussions/1324

AdrianGarside: Sounds overly complicated but it would be great to have an actual way to express this kind of state.

AdrianGarside (approved):

PR #2068, opened 2026-08-05
fix: include AMS HT filament in print weight and length attributes

### LINKED ISSUE #2029: [Bug] Print Weight sensor does not include AMS HT filament weight
### Home Assistant Version

2026.1.1

V2.2.22

### Describe the bug

Description

The print_weight sensor does not include the filament weight assigned to the AMS HT.

When a print uses filament from the AMS 2 Pro, the print_weight sensor correctly provides an attribute showing which AMS tray the filament weight belongs to.

For example:

AMS 1 Tray 1: 25.6

However, when a print uses filament from the AMS HT, no corresponding AMS HT attribute is added to the print_weight sensor.

As a result, Home Assistant only provides the print weight, but there is no information indicating that this weight belongs to the AMS HT.

Expected behavior

When filament from the AMS HT is used, the print_weight sensor should provide a separate attribute for the AMS HT, similar to the existing AMS 2 Pro attributes.

For example:

AMS 1 Tray 1: 25.6
AMS HT Tray 1: 18.3

The exact attribute name is not important. The important part is that the filament weight used from the AMS HT can be identified separately.

Actual behavior

Filament weight from the AMS 2 Pro is correctly shown in the print_weight attributes:

AMS 1 Tray 1: 25.6

Filament weight from the AMS HT has no corresponding attribute.

Therefore, it is not possible to determine from the print_weight sensor how much filament was used from the AMS HT.

Steps to reproduce

1. Connect an AMS 2 Pro and an AMS HT to the printer.
2. Start a print using filament from an AMS 2 Pro tray.
3. Check the attributes of the print_weight sensor.
4. The filament weight is correctly shown under an attribute such as AMS 1 Tray 1.
5. Start a print using filament from the AMS HT.
6. Check the attributes of the same print_weight sensor.
7. There is no attribute for the AMS HT filament weight.

Why this matters

I use the print_weight sensor in Home Assistant and Node-RED to record filament consumption and deduct the used weight from the corresponding spool in Spoolman.

The AMS 2 Pro can be handled correctly because the sensor provides attributes such as AMS 1 Tray 1.

The AMS HT cannot be handled because its filament weight is not identified in the sensor attributes. This prevents the automation from determining which spool should have its remaining weight deducted.

Hardware

* Printer: Bambu Lab H2C
* Filament systems:
    * AMS 2 Pro
    * AMS HT

Additional information

This issue appears to affect the AMS HT mapping in the print_weight sensor.

The AMS 2 Pro tray weight is exposed correctly, but the AMS HT weight is missing.

I can provide Home Assistant diagnostics, entity attribute screenshots, MQTT data, or print files if required.

### To Reproduce

h2c use ams 2pro and ht to print,see the print_weight data

### Expected Behaviour

Image

H2C

### Diagnostic Output

_No response_

### Log Extracts

_No response_

### Other Information

Image
total weight is 5.33,ams 2pro  AMS 2 Tray 1 is 2.26,but not ams ht weight

Oose97:
## Description

The `print_weight` and `print_length` sensors omit any filament drawn from an AMS HT. A print that uses one reports a correct total but no matching per-slot attribute, so automations cannot tell which spool to deduct — the Spoolman case in #2029.

Filament sources are numbered two different ways, and the two were being conflated:

| Scheme | Regular AMS | AMS HT |
|---|---|---|
| **Slot index** — `print.ams_mapping`, cloud `amsDetailMapping.ams` | 0-15, as `unit * 4 + tray` | **16-23** |
| **Unit id** — module names (`n3s/128`), `amsDetailMapping.amsId`, active tray | 0-3 | 128-135 |

Captured from a live two-colour print using one AMS 2 Pro tray and one AMS HT. The `amsDetailMapping` entry pairs the two schemes explicitly, showing flat slot `16` is unit `128`:

```
ams_mapping: [3, 16]

ams_print_data: [{'ams': 3,  'weight': 1.18,  'amsId': 0,   'slotId': 3},
                 {'ams': 16, 'weight': 24.88, 'amsId': 128, 'slotId': 0}]
```

Two separate defects followed from this:

1. **`get_print_weights` / `get_print_lengths` only scanned slots 0-15** (`for i in range(16)`), so an AMS HT slot could never be reported — even when its weight had already been recorded correctly by the cloud task parser, which bounds-checked against the full 136-entry array.

2. **The 3mf parser gated on `ams_index < 16`**, a line already carrying `# BUG - This will not yet handle AMS HT devices`. The filament was discarded and fell through to the default external-spool label:

```
AMS 1 Tray 4:   0.45m |  1.38g
External spool: 16.06m | 49.43g     <- actually the AMS HT
```

### Fix

Adds `ams_slot_name()` as the single owner of the slot numbering and routes all four call sites through it: both attribute builders, the 3mf parser and the cloud task parser. This keeps weights and lengths from drifting apart again. The magic `136` array size is replaced with named constants derived from the two schemes.

AMS HT slots are reported as `AMS HT 1` … `AMS HT 8`, with no tray number, since an AMS HT holds a single spool. Existing `AMS n Tray m` names are unchanged, so current automations keep working.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### Link to Issue

Fixes #2029

## Documentation

## Testing

Unit tests added to `test_models.py`, 32 pass:

- `test_ams_slot_name` — regular slots 0-15, AMS HT slots 16-23, AMS HT unit ids 128-135, and rejection of 24 / 127 / 136 / 255 / -1
- `test_get_print_weights_includes_ams_ht` and `test_get_print_lengths_includes_ams_ht` — built from the real payload above

Verified on hardware: P2S in LAN mode with an AMS 2 Pro and an AMS HT, two-colour print drawing one filament from each.

Before, with debug logging enabled to capture the payloads quoted above:

```
sensor.print_weight = 50.82 g    attributes: AMS 1 Tray 4: 1.38
sensor.print_length = 16.51 m    attributes: AMS 1 Tray 4: 0.45
```

The AMS HT's filament was absent from both, and the 3mf parser logged it as `External spool: 16.06m | 49.43g`.

After, same print:

```
sensor.print_weight = 50.82 g    attributes: AMS 1 Tray 4: 1.38,  AMS HT 1: 49.43
sensor.print_length = 16.51 m    attributes: AMS 1 Tray 4: 0.45,  AMS HT 1: 16.06
```

The per-slot values now sum to the totals, and the `16.06 m / 49.43 g` previously mislabelled as the external spool is attributed to `AMS HT 1`.

A single-filament print from the AMS HT was also checked, since that path was broken too:

```
sensor.print_weight = 83.51 g    attributes: AMS HT 1: 83.51
sensor.print_length = 27.12 m    attributes: AMS HT 1: 27.12
```

Regular AMS trays continue to report unchanged, as `AMS n Tray m`.

## Additional Notes

- Single-filament AMS HT prints were affected too, not just multi-material ones. `ams_mapping` is `[16]`, so the attribute was missing there as well; only the total looked correct, because it comes from the plate's `weight` key and never passes through the mapping. Both cases are covered in the testing above.
- The unit-id range 128-135 is accepted in addition to the slot range because that numbering appears on AMS identifiers elsewhere in the codebase. Only 16-23 was observed in `ams_mapping` and `amsDetailMapping`, so the extra range is defensive; happy to drop it if you would rather the helper accepted only the slot space.
- Unrelated, but noticed while debugging: `BambuOptionsFlowHandler` snapshots the logger level in `__init__` and restores it in `__del__`, so opening the options dialog can silently reset a level set via `logger.set_level`, at whatever later moment garbage collection runs. Happy to raise that separately.


AdrianGarside:
"at whatever later moment garbage collection runs" - yeah I couldn't find a more reliable way to restore it at the right time. In practice it was good enough. But if there's a way to reliably run it more immediately when the dialog goes away that would be nice.

PR #2069, opened 2026-08-06
fix: MQTT hot-path bugs — disconnect stall, blocking cloud fetch, shared command templates

it-rec:
## Description

Seven fixes in the MQTT/coordinator hot path and in `pybambu`. Each one is visible in the code rather than inferred from a benchmark, so they are listed with the mechanism.

**`disconnect()` blocked the event loop for a guaranteed 5 s.** The MQTT thread was joined with `join(timeout=5)` *before* the paho network loop was broken. `loop_forever()` only returns once `disconnect()` has been called, so the join could never succeed early — every reload and every shutdown paid the full timeout inside Home Assistant's event loop. The loop is now stopped first, then the thread is joined.

**Slicer settings were fetched over blocking cloud HTTP on the paho network thread.** `on_message` called `slicer_settings.update()` directly while `keepalive` was 5 s. A slow cloud response stalled the keepalive ping, the broker dropped the connection, and the client went into a reconnect loop. The fetch now runs on its own daemon thread, and `keepalive` is raised to 30 s — the watchdog still detects genuinely dead connections after 60 s.

**Command templates were mutated in place as shared module-level dicts** (gcode, speed profile, upgrade confirm, AMS commands, `print_project_file`, `skip_objects`). Two concurrent service calls could corrupt each other's payloads, and `upgrade_confirm` baked the first printer's firmware URL into the global and resent it for later upgrades. All are deep-copied before mutation now.

**`upgrade_template()` raised `AttributeError` on an unparseable firmware URL** instead of returning `None`.

**Cloud task `endTime` was written into `start_time`** instead of `end_time`, corrupting the start-time sensor and the usage-hours accounting.

**The change-detection result was computed and then discarded.** `print_update()` already returns whether a submodule reported a change, but `event_printer_data_update` fired unconditionally on every push — roughly once per second — so all entities recomputed and rewrote their state even when nothing had changed. The existing return value is now honoured. Behaviour on first connect is unchanged: the initial full push always reports changes, and `event_printer_ready` still fires separately.

**Two events were fired but never handled.** `event_printer_pick_image_update` meant the pick image entity never updated; `event_printer_missing_sdcard` is now exposed as a device trigger ("Missing SD card", en + de), and `PublishDeviceTriggerEvent` no longer raises when the device is not yet registered.

Camera frame updates (chamber/cover/pick) are also dispatched straight to the affected image entity instead of triggering a full coordinator refresh per frame.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Other (please describe): reduced per-message work in the coordinator

### Link to Issue

N/A — found while reading the hot path, not from a filed report.

## Documentation

## Testing

The existing `pybambu` unit tests pass. Beyond that, tested on a **P2S with AMS 2 Pro in LAN-only mode**, running on Home Assistant on an ARM host: the integration loads cleanly, 69 entities, no errors in the log.

**On the CPU claim, I want to be straight with you: I could not measure it, and I am not asserting a number.** I ran paired 6-minute container-CPU samples with the printer idle-connected, old build vs new. The apparent gain (≈42.6 → ≈39.9 CPU-s) is real in the data but the same magnitude shows up *with the opposite sign* in the control condition where the printer is powered off and no status pushes exist at all — a difference cannot exist there by construction, so that control is not a control, and the measurement does not separate signal from noise. Instantaneous container CPU on this host swings ~8 points between samples; the effect being chased is under 1.

Counting `state_changed` events does not work as a proxy either: Home Assistant suppresses writes that do not change a value, so the redundant recomputation cost CPU without ever producing an event. Both builds sit at 0–5 Bambu events per minute.

So the gating change rests on the code, not on a benchmark: the change-detection result was already being computed and thrown away. If you want it benchmarked properly it would need a counter inside `_update_data()` rather than an inference from process CPU — happy to add one if you would rather see that before merging.

The bug fixes above are all reviewable by inspection. The ones I could not exercise on my hardware are `upgrade_template()` (no unparseable firmware URL to hand), the cloud `end_time` path (LAN-only setup), and the concurrent-template corruption (verified by reading, not by racing two service calls).

## Additional Notes

Happy to split this into separate PRs if you would prefer to take the `disconnect()` and slicer-settings threading fixes on their own — those two are the ones with user-visible impact.

### DIFF
--- a/custom_components/bambu_lab/coordinator.py
+++ b/custom_components/bambu_lab/coordinator.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import copy
 import functools
 import os
 import re
@@ -15,6 +16,7 @@ from homeassistant.helpers import (
     device_registry,
     entity_registry
 )
+from homeassistant.helpers.dispatcher import async_dispatcher_send
 from homeassistant.helpers.entity import DeviceInfo
 from homeassistant.helpers.update_coordinator import DataUpdateCoordinator
 from homeassistant.core import Event, HomeAssistant, callback
@@ -148,12 +150,21 @@ class BambuDataUpdateCoordinator(DataUpdateCoordinator):
                     data=self.config_entry.data,
                     options=options)
 
+        # Image updates are dispatched directly to the affected image entity instead of
+        # refreshing the full entity set - chamber images arrive at up to a couple of
+        # frames per second and a full refresh per frame is needlessly expensive.
         elif event == "event_printer_chamber_image_update":
             if self.get_option_enabled(Options.IMAGECAMERA):
-                self._update_data()
+                async_dispatcher_send(self._hass, self.image_update_signal("chamber"))
 
         elif event == "event_printer_cover_image_update":
-            self._update_data()
+            async_dispatcher_send(self._hass, self.image_update_signal("cover"))
+
+        elif event == "event_printer_pick_image_update":
+            async_dispatcher_send(self._hass, self.image_update_signal("pick"))
+
+        elif event == "event_printer_missing_sdcard":
+            self.PublishDeviceTriggerEvent(event)
 
         elif event == "event_printer_error":
             self._update_printer_error()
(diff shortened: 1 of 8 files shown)


AdrianGarside:
Some nice quality improvements there. Thanks.

PR #2074, opened 2026-08-07
Refresh model when delayed task name arrives

### LINKED ISSUE #1804: [Bug] The cover image is often wrong when reprinting from the printer.
### Home Assistant Version

2025.12.4

v2.2.18

### Describe the bug

Cover image is of something old I printed, this happens a lot, the task name is correct but the image of something I printed in the past.
Image

### To Reproduce

I'm not sure how to reproduce it, I keep printing different things and sometimes the cover image isn't correct even though task name is.

### Expected Behaviour

Cover image should match what is being printed.

X1C

### Diagnostic Output

_No response_

### Log Extracts

_No response_

### Other Information

_No response_

Ali229:
## What changed

- Re-resolve print model data when a printer-initiated print or reprint reports its `subtask_name` after model data was already loaded.
- Limit the refresh to active prints and the first empty-to-named transition.
- Add regression tests for delayed task names, repeated task-name updates, and finished prints.

## Root cause

Some printer-initiated prints reach `RUNNING` before MQTT supplies `subtask_name`. The integration then falls back to the newest `.3mf` on the printer, which can be a previous slicer upload. Because model data is marked loaded, the later correct task name did not trigger another lookup, leaving a stale cover image and metadata.

## Impact

When the delayed task name arrives, the integration now clears the fallback model data and resolves the named `.3mf`, allowing the cover image and metadata to match the active print.

Fixes #1804

## Validation

- `python test_models.py TestPrintJob -v` — 4 passed
- Broader direct discovery — 58 tests passed; 2 unrelated modules were not discoverable because they use package-relative imports under script-style discovery
- `python -m compileall -q ..`
- `git diff --check`

### DIFF
--- a/custom_components/bambu_lab/pybambu/models.py
+++ b/custom_components/bambu_lab/pybambu/models.py
@@ -1030,6 +1030,21 @@ class PrintJob:
         self._subtask_name = data.get("subtask_name", self._subtask_name)
         if old_subtask_name != self._subtask_name:
             LOGGER.debug(f"SUBTASK_NAME: {self._subtask_name}")
+
+        # Printer-initiated prints and reprints can reach RUNNING before the
+        # printer reports a subtask name. In that case model data is initially
+        # loaded using the newest 3mf on the printer as a fallback, which may
+        # belong to an older print. Resolve the model again once the real task
+        # name arrives so the cover image and metadata match the active print.
+        if (
+            old_subtask_name == ""
+            and self._subtask_name != ""
+            and self._loaded_model_data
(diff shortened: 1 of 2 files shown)


Ali229:
@greghesp  Hi Greg, this issue has been around for quite a while (#1804), and I put together a fix in this PR along with regression tests. When you get a chance, could you or one of the maintainers take a look and see if it can be merged?

If you’re looking for additional help maintaining the project, I’d also be happy to contribute more regularly and be added as a collaborator. Thanks.

PR #2082, opened 2026-08-23
ci: run pybambu unit tests on every PR

greghesp:
## Description

The `pybambu` library has a real unit test suite (63 tests across 5 files) but nothing in CI ever ran it — there's no `pytest` job, only HACS/hassfest validation. This PR wires it up.

It's not just adding a workflow file, though: `pybambu/tests/` lived *inside* `custom_components/bambu_lab/`, whose `__init__.py` imports `homeassistant`. Pytest's package collector walks every `__init__.py`-bearing ancestor directory between rootdir and a test file and imports it as part of collection — so collecting anything under the old `pybambu/tests/` path forced an import of the whole integration (and transitively `config_flow.py`, `ssdp`, `mqtt`, `http`, etc.), none of which belong in a fast unit-test job and aren't installed here.

**Changes:**

- Move the test suite + fixtures to a new top-level `tests/pybambu/` (sibling of `custom_components/`), so pytest never needs to import `custom_components/bambu_lab/__init__.py` to reach them. No `__init__.py` in that directory on purpose — giving it one would make pytest bind the name `pybambu` to the test package instead of the real library, shadowing it.
- Extract `MockMQTTClient` (+ its `MqttMessageInfo`/message helper types) out of `pybambu/tests/test_utils.py` into `pybambu/mock_mqtt_client.py`. It was production code masquerading as test code — `bambu_client.py` imports it for the `MOCK-<model>` demo-serial feature — which meant every release zip shipped the entire test suite and fixtures to end users just to satisfy that import. The extracted version also drops a `unittest.mock.MagicMock` stand-in for MQTT messages in favour of a small dedicated class, since production code has no business depending on `unittest.mock`.
- Move the `MOCK-*.json` payload fixtures alongside it to `pybambu/mock_data/`, since `MockMQTTClient` loads them at runtime.
- Add `tests/conftest.py` (puts `custom_components/bambu_lab` on `sys.path`) and `tests/requirements.txt` (pytest + pybambu's own runtime deps — no `homeassistant` needed).
- Replace `pybambu/run_tests.sh` (a local-only script needing a hand-built venv) with `tests/run_tests.sh`, a thin wrapper around the same suite CI now runs.
- New `.github/workflows/tests.yml`: runs `pytest tests/pybambu` on every push/PR against Python 3.13.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Other (please describe): CI + test-suite reorganization, no behavior change to the integration itself

### Link to Issue

None — this came out of a repo/architecture review, not a filed issue.

## Documentation

## Testing

Verified locally, not just "tests pass green":
- All 63 tests pass in a clean venv installing only `tests/requirements.txt` — no `homeassistant`.
- Every `.py` file under `custom_components/bambu_lab/` still compiles.
- The CSRF regression tests (`test_bambu_cloud_csrf.py`) were run with `socket.connect`/`socket.create_connection`/`requests.get`/`requests.post` monkeypatched to hard-fail — confirming they're pure logic tests with zero network dependency, safe to run unauthenticated on GitHub's runners.
- The `MockMQTTClient` extraction was exercised end-to-end outside of pytest: publish/subscribe/`on_message` round-trip against `MOCK-A1.json`, then fed through the real `pybambu.models.Info` parser, confirming the mock-printer demo feature behaves identically after the move.

## Additional Notes

No changes to `coordinator.py`, entity platforms, or `config_flow.py` — this is scoped entirely to test infrastructure and the one production file (`bambu_client.py`) that needed its import path updated.


AdrianGarside:
@greghesp This is a nice change but other PRs added new tests so it's got merge conflicts now.

AdrianGarside:
@greghesp is this ready to go in?

greghesp:
> @greghesp is this ready to go in?

Yep, should be good to go

PR #2093, opened 2026-08-27
Fix RTSP URL validation and camera availability

FiraSenax:
## Summary

- reject disabled, missing, or malformed RTSP endpoints instead of exposing them as camera streams
- safely construct authenticated RTSP URLs, including URL-escaped access codes and configured-host substitution
- report the camera unavailable until the printer supplies a usable RTSP endpoint
- preserve all existing AMS device names and identifiers

## Compatibility

- no device or entity naming changes
- no model-support or startup-registration fallback changes
- no behavior change for image-based cameras
- RTSP handling remains behind the existing `Features.CAMERA_RTSP` check

## Validation

- rebased by merging the latest `main` and adopting the new `tests/pybambu` layout
- `pytest tests/pybambu -v`: 83 passed
- `git diff --check`
- verified against the H2C payload value `disable` observed while LAN LiveView was disabled


custom_components/bambu_lab/pybambu/models.py
```diff
@@ -175,7 +176,7 @@ def supports_feature(self, feature):
         # processing the mqtt payload and so may be called before full initialization is complete as it processes
         # the very first payload.
         if feature == Features.CAMERA_RTSP:
-            return model in (h2_printers | p2_printers | x1_printer | x1e_printer | x2_printers)
+            return model in RTSP_CAMERA_PRINTERS
```
AdrianGarside: Match the existing local only variable above (lines 163-171) instead of a global const. I see you did this to use the list elsewhere but I don't currently agree with that change being necessary or appropriate.

custom_components/bambu_lab/camera.py
```diff
@@ -59,7 +59,14 @@ async def async_setup_entry(
...
 
     if coordinator.get_option_enabled(Options.CAMERA):
-        if coordinator.get_model().supports_feature(Features.CAMERA_RTSP):
+        configured_model = entry.data.get("device_type")
+        try:
+            configured_model = Printers(configured_model)
+        except ValueError:
+            configured_model = None
+
+        if (coordinator.get_model().supports_feature(Features.CAMERA_RTSP)
```
AdrianGarside: Unclear why this change is needed. There shouldn't be any need to have this fallback since we don't initialize things until we see the printer payload. I want to understand why you think this change is necessary.

custom_components/bambu_lab/pybambu/utils.py
```diff
@@ -39,6 +40,39 @@ def search(lst, predicate, default={}):
     return default
 
 
+def get_ams_unit_name(index: int, model: str) -> str:
+    """Return a human-readable AMS unit name without exposing protocol indices."""
+    if model == "AMS HT" and 128 <= index <= 135:
+        return f"AMS HT {index - 127}"
```
AdrianGarside: This is a breaking change for a nice to have. Remove this.

AdrianGarside:
You'll also need to rebase this one due to the tests getting moved out of the product code.

FiraSenax:
Resolved against the latest `main` and the new `tests/pybambu` layout. I also addressed the earlier review: the AMS naming change, global RTSP model set, and startup fallback are removed. The PR is now limited to RTSP URL validation/camera availability plus tests; all 83 tests pass.

PR #2094, opened 2026-08-27
Add H2C Vortek hotend usage sensors

FiraSenax:
## Summary

- expose the H2C Vortek `p_t` telemetry as six per-hotend usage sensors
- report usage in hours with the duration device class and total-increasing state class
- keep all new sensor creation behind the existing `Features.HOTEND_RACK` gate
- preserve all existing external-spool device names and identifiers

## Why

Real H2C telemetry reports `p_t` for rack nozzle IDs 16-21. Bambu Studio parses this field as nozzle print time (`m_nozzle_print_time`) and displays it as used time. Keeping it as a dedicated numeric sensor lets Home Assistant retain history and statistics.

## Compatibility

- This is additive: existing device names and identifiers are unchanged.
- Vortek sensors are created only when `Features.HOTEND_RACK` is supported.
- Non-Vortek nozzle payloads are ignored by the model.
- Empty rack slots return no usage value rather than a false zero.
- H2D behavior remains unchanged.

## Validation

- `custom_components/bambu_lab/pybambu/run_tests.sh`: 52 passed
- `git diff --check`
- payload shape verified against a real idle H2C with six Vortek slots
- live validation performed during an H2C print using the left nozzle; monitoring was read-only

## Upstream protocol reference

- Bambu Studio parses `p_t` into `m_nozzle_print_time`: https://github.com/bambulab/BambuStudio/blob/master/src/slic3r/GUI/DeviceCore/DevNozzleSystem.cpp
- Bambu Studio presents it as nozzle used time: https://github.com/bambulab/BambuStudio/blob/master/src/slic3r/GUI/Widgets/NozzleRack/wgtDeviceNozzleRackUpdate.cpp


FiraSenax:
Live H2C telemetry validation completed on HA Core 2026.8.3 (read-only; printer idle, no heat/motion/print commands):

- rack slots 2-6 reported `p_t` directly under nozzle IDs 17-21
- the mounted slot 1 hotend reported under extruder ID 0 with `tar_id` 16
- `HotendRack.print_update()` correctly maps that mounted extruder entry back to rack slot 16
- all six values converted to plausible hour totals using `p_t / 3600`
- non-Vortek and H2D regression tests remain green

The production HA installation was not overwritten: HACS rejected installing a raw PR commit/branch, so the temporary custom-repository/test-release metadata was removed and v2.2.25 remains installed.

FiraSenax:
Additional real-print validation on an H2C (read-only monitoring, first two layers):

- left nozzle was selected for the job; HA generic nozzle temperature/target correctly followed the left nozzle while the right target remained 0 °C
- rack slot 16 remained mounted throughout preparation, nozzle cleaning, bed leveling, heating, and layers 1-2
- holder stayed centered/idle and no false rack transition occurred
- AMS HT became active and the active tray resolved correctly even though raw `tray_now` temporarily alternated between 0 and 255 during preparation
- `p_t` was stable during the observed first layers, indicating the firmware publishes this cumulative value on a slower/full-status cycle rather than continuously; this is appropriate for a total-increasing diagnostic sensor
- no HMS or print errors were reported

The print itself was not controlled or modified.

AdrianGarside:
@FiraSenax the device name change is a breaking change - it will break the cards and it will break peoples dashboards if they have references to the devices. You've also only change it for the H2C and not the H2D so now it's inconsistent. I'd rather not break people just for a nice to have name change.

FiraSenax:
Agreed. I removed the H2C external-spool display-name change and restored the original naming path for every printer. The PR is now additive only: it adds the H2C Vortek used-time sensors behind `Features.HOTEND_RACK`; existing H2C/H2D device names and identifiers remain unchanged. The pybambu suite passes (52 tests).

FiraSenax:
That makes sense — we should not risk breaking existing cards and dashboards for a cosmetic device-name change. I have removed the spool naming change and pushed the cleanup; the PR now only contains the additive H2C Vortek used-time sensors.

AdrianGarside:
@FiraSenax the tests got moved out of the product code so this PR has conflicts now. Please can you resolve.

FiraSenax:
Resolved by merging the latest `main`. The H2C regression tests are now under `tests/pybambu/test_models.py`, and the updated suite passes: 83 tests.

Issue #2065, opened 2026-08-03
Unable to authenticate


waspsoton:
# Unable to authenticate after previously working

## Describe the bug

The integration has been working without any issues for a long time, including the camera feed. I recently noticed that the camera was no longer updating.

After attempting to reconfigure the integration, I can no longer authenticate. The setup fails with the following error:

> **Failed to connect. Check credentials.**

I have verified that the credentials are correct because I can successfully sign in to MakerWorld using the same username and password.

## To Reproduce

1. Open the Bambu Lab integration.
2. Attempt to sign in using valid MakerWorld credentials.
3. Authentication fails with **"Failed to connect. Check credentials."**

## Expected behavior

The integration should authenticate successfully and reconnect to the printer.

## Environment

* Home Assistant version: **2026.7.4**
* Installation type: **Home Assistant OS**
* Integration: **https://github.com/greghesp/ha-bambulab**

## Additional context

* The integration was previously working correctly.
* The camera feed stopped working first.
* After trying to reconnect the integration, authentication now always fails.
* I can log into MakerWorld with the same credentials, so they appear to be valid.

Please let me know if there are any logs or debug information I can provide.

waspsoton:
Just in case it helps anyone else:

I was able to get this working by installing the current `main` branch instead of the latest tagged beta release.

Steps I took:

* Backed up my existing `bambu_lab` integration.
* Installed the current `main` branch from GitHub (`manifest.json` reports version `0.0.0`).
* Removed my backup integration from `/config/custom_components` so Home Assistant wasn't loading it as a second custom integration.
* Restarted Home Assistant.
* Added the integration again and completed the login flow.

The login now succeeds.

The key thing that caught me out was that my backup folder (`bambu_lab_backup`) was still inside `/config/custom_components`, so Home Assistant was loading it as a custom integration as well.

Hopefully this helps someone else who runs into the same issue.

luketenu:
All my entities became unavailable recently. I tried re-configuring once with selecting my email address and once with selecting Bambu Lab cloud but both failed.

<img width="1440" height="2122" alt="Image" src="https://github.com/user-attachments/assets/a096b513-3936-4e15-9388-ba70c67c6898" />
<img width="1440" height="1229" alt="Image" src="https://github.com/user-attachments/assets/493b5501-02ea-4268-b7c7-564479211703" />

caliban511:
我这边也无法验证，hass不停的报认证失败。

[...]

Issue #1863, opened 2026-01-29
[Feature] Reduce log spam (Warning Error Code 5) when printer is powered off


thib3113:
### Describe the feature

The integration currently logs a `WARNING ... On Disconnect: Printer disconnected with error code: 5` every 5 seconds when the printer is powered off (e.g., via a smart plug). This creates significant log spam in Home Assistant.

I would like to request a way to mitigate this, such as:
1. **Log Silencing / Exponential Backoff for Logs:** Instead of spamming the same warning every 5 seconds, could the integration suppress repeated identical warnings or increase the log interval significantly after the first few failures? This would keep the logs clean while still signaling the initial disconnection.
2. **Integration "Standby" or "Off" Switch:** Would it be possible to implement a service or a switch to tell the integration "the printer is off"? 
   - This would stop all reconnection attempts (and thus all logs) until the printer is marked as "on" again.
   - Alternatively, if the integration could detect the "off" state (after X failed attempts), it could transition to a "Standby" mode with very infrequent polling.

### What device is this for?

A1

### Other Information

While I could manually set the log level to `error` in my `configuration.yaml`, I am concerned about missing "real" critical errors if I silence the entire component. 

Is it safe to mask all warnings for this integration? If not, a more granular way to handle "printer unreachable" states would be a great QoL improvement.

**Example Log Spam:**
2026-01-29 06:23:10.177 WARNING (A1-Mqtt-xxx) [custom_components.bambu_lab.pybambu] On Disconnect: Printer disconnected with error code: 5
2026-01-29 06:23:15.009 WARNING (A1-Mqtt-xxx) [custom_components.bambu_lab.pybambu] On Disconnect: Printer disconnected with error code: 5

[config_entry-bambu_lab-01JD45VNNK1THWZM13K6QWWTNT.json](https://github.com/user-attachments/files/24957667/config_entry-bambu_lab-01JD45VNNK1THWZM13K6QWWTNT.json)

AdrianGarside:
Can you include a diagnostics log too so I can see how you have the integration configured to connect to the printer.

thib3113:
@AdrianGarside added

zygote279:
Yes, I'm getting this too - eg 13,000+ occurrences! :

Log details (WARNING)
This error originated from a custom integration.
Logger: custom_components.bambu_lab.pybambu
Source: custom_components/bambu_lab/pybambu/bambu_client.py:533
[...]

PR #2110, opened 2026-08-31
Recover automatically when the printer rejects the access code

jobshimo:
## Description

When a printer rejects the MQTT connection with CONNACK code 5 (Access Denied), the integration currently logs one error and stops the client (`bambu_client.py`, `on_disconnect`) — and that is the end of it: no repair issue, no reauth path, no recovery. The entities go `unavailable` and the only explanation lives in a debug log.

The usual real-world cause is a **rotated LAN access code**: a factory reset regenerates it (other causes: the user regenerating it manually). After the reset, the user re-binds the printer to their Bambu account — Bambu's own required step — and from that moment **Bambu Cloud already knows the new access code**, which is how Bambu Studio and Handy reconnect without asking for anything. The integration holds the same cloud session and the same API (`get_device_list()` returns `dev_access_code`), but never uses it for this.

This PR closes that gap:

- `pybambu/bambu_client.py`: on the first CONNACK 5 of an episode, emit a new `event_printer_access_denied` callback (additive; existing behaviour unchanged).
- `coordinator.py`: handle the event once per failure episode (reset on `event_printer_ready`):
  - **Cloud-linked printers**: fetch the device list from Bambu Cloud (in an executor) and compare `dev_access_code` with the configured one. If it differs, update the config entry and reload — the integration heals itself with zero user action, matching the behaviour of Bambu's own apps.
  - If the cloud reports the **same** code, do nothing: code 5 is also seen when the printer is powered off (#1863), and replacing a correct code or nagging the user would be wrong there.
  - **LAN-only setups** (or a failed cloud lookup): raise a visible repair issue (`access_code_rejected`) telling the user to read the new code off the printer's screen — instead of today's silent dead integration.

Translations: `access_code_rejected` issue strings added to `en.json` and `es.json` (other locales left to the informational parity flow).

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### Link to Issue

No exact open issue; related context: #2065 (authentication dead-end), #1863 (code 5 while the printer is powered off — the reason the fix compares codes instead of reacting blindly).

## Documentation

## Testing

`python -m pytest tests/pybambu`: **89 passed** (86 existing + 3 new in `tests/pybambu/test_bambu_client_disconnect.py` pinning that CONNACK 5 fires the event once per episode and other codes do not). Both translation files validate as JSON; touched modules compile clean.

Field validation: the cloud-recovery path reproduces a real incident — after a factory reset, an integration configured via cloud login sat in an Access Denied loop; fetching `dev_access_code` via the stored session and updating the entry's `access_code`, then reloading, restored the connection immediately (CONNACK 0). This PR automates exactly that manual repair.

## Additional Notes

- The recovery runs at most once per failure episode (`_access_denied_handled`, reset on a successful connect), so a printer that stays denied does not hammer the Bambu API.
- The stored code is only ever **replaced** when the cloud reports a genuinely different one — never cleared, never guessed.
- Scope deliberately excludes the reconnect/backoff behaviour for other disconnect codes.

### DIFF
--- a/custom_components/bambu_lab/coordinator.py
+++ b/custom_components/bambu_lab/coordinator.py
@@ -40,6 +40,7 @@ from .const import (
 )
 
 from .pybambu import BambuClient
+from .pybambu.bambu_cloud import BambuCloud
 from .pybambu.const import (
     AMS_MODELS,
     AMS_DRYING_MODELS,
@@ -82,6 +83,9 @@ class BambuDataUpdateCoordinator(DataUpdateCoordinator):
             
         self._updatedDevice = False
         self._shutdown = False
+        # Guards the access denied recovery so one failure episode triggers at most one
+        # cloud lookup. Reset once the printer connects successfully again.
+        self._access_denied_handled = False
         self.data = self.get_model()
         self._eventloop = asyncio.get_running_loop()
         # Pass LOGGERFORHA logger into HA as otherwise it generates a debug output line every single time we tell it we have an update
@@ -116,7 +120,10 @@ class BambuDataUpdateCoordinator(DataUpdateCoordinator):
         
         if event == "event_printer_bambu_authentication_failed":
             self._report_authentication_issue()
-        
+
+        elif event == "event_printer_access_denied":
+            self._hass.async_create_task(self._async_handle_access_denied())
+
         elif event == "event_printer_no_external_storage":
             self._report_no_external_storage_issue()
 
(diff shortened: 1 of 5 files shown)


AdrianGarside:
'usual' is probably a big overstatement given the frequency with which I see access denied myself and in issues where the access code isn't incorrect but the change is still good :)

Issue #2089, opened 2026-08-26
[Bug] AMS device wrongly removed as "stale" by _printer_ready() race condition on restart


Mukkemann1972:
## Title
[Bug] AMS device wrongly removed as "stale" by `_printer_ready()` race condition on restart

## Body

**Describe the bug**
On a Home Assistant restart, the AMS device attached to my printer occasionally gets permanently
removed from the device registry (device + all its entities gone), even though the AMS is
correctly attached and reporting normally moments later.

**Root cause (traced in the integration source, `coordinator.py`)**
`_printer_ready()` runs an "orphaned AMS" cleanup on every connect:

```python
def _printer_ready(self):
    ...
    # First clean up orphaned AMS devices such as when an AMS is moved between printers.
    existing_ams_devices = []
    for index in self.get_model().ams.data.keys():
        ams_entry = self.get_model().ams.data[index]
        if ams_entry is not None:
            existing_ams_devices.append(ams_entry.serial)
    ...
    for device in dev_reg.devices.values():
        if config_entry_id in device.config_entries:
            if device.model == 'AMS' or device.model == 'AMS Lite' or device.model == 'AMS 2 Pro' or device.model == 'AMS HT':
                ams_serial = list(device.identifiers)[0][1]
                if ams_serial not in existing_ams_devices:
                    LOGGER.debug(f"Found stale attached AMS with serial {ams_serial}")
                    ams_devices_to_remove.append(device.id)

    for device in ams_devices_to_remove:
        LOGGER.debug("Removing stale AMS.")
        dev_reg.async_remove_device(device)
```

This compares the registry against `self.get_model().ams.data`, which is populated by parsing
live MQTT `print_update` payloads. `_printer_ready()` itself fires as soon as the printer signals
"ready" — which can happen **before** a full status payload containing AMS data has actually been
received and parsed into `ams.data`. If that race loses, `ams.data` is still empty at the moment
this cleanup runs, so every currently-attached AMS gets flagged "stale" and removed — permanently,
since this pass only runs once per config-entry setup and won't retry.

**To Reproduce**
Timing-dependent, but reliably hit this on a real instance:
1. Have a printer with an attached AMS already bound and working.
2. Restart Home Assistant (a cold reconnect makes the race more likely than a warm reload).
3. Check Settings → Devices & Services → Bambu Lab: the AMS sub-device is sometimes gone
   (device + all entities), depending on how quickly the first full MQTT payload with AMS data
[...]

PR #2113, opened 2026-09-02
Fix AMS entities disappearing after incomplete version data

Ali229:
## Summary

Fixes a startup race where AMS entities can disappear while the printer itself remains online.

## Root cause

A `push_status` payload can arrive before complete `get_version` metadata. In that order, pybambu creates a placeholder AMS with model `Unknown` and an empty serial. Home Assistant can then initialize against that placeholder before the real AMS identity arrives.

If a later `get_version` response fills in the AMS serial/model, the integration previously did not rerun entity setup. This can leave an AMS device registered with no entities until the integration is manually reloaded.

## Changes

- Do not register AMS sensor/binary-sensor entities until the AMS has a real serial.
- Make `AMSList.info_update()` report AMS identity changes.
- Retrigger the existing printer-ready reinitialization path when late version metadata fills in AMS identity after startup.
- Remove stale placeholder AMS devices created with an empty identifier by older behavior.
- Add regression coverage for `push_status` arriving before incomplete then complete `get_version` data.

## Tests

- `python -m pytest -q` — 95 passed
- `git diff --check upstream/main...HEAD` — clean

### DIFF
--- a/custom_components/bambu_lab/sensor.py
+++ b/custom_components/bambu_lab/sensor.py
@@ -45,6 +45,13 @@ async def async_setup_entry(
 
     for sensor in AMS_SENSORS:
         for index in coordinator.get_model().ams.data.keys():
+            ams_instance = coordinator.get_model().ams.data[index]
+            # push_status can arrive before complete get_version metadata and create
+            # a placeholder AMS with no serial. Device/entity identity depends on the
+            # real AMS serial, so wait for version metadata instead of registering a
+            # transient placeholder that HA cannot later reconcile cleanly.
+            if ams_instance is None or not ams_instance.serial:
+                continue
             if sensor.exists_fn(coordinator, index):
                 async_add_entities([BambuLabAMSSensor(coordinator, sensor, index)])
 
--- a/custom_components/bambu_lab/binary_sensor.py
+++ b/custom_components/bambu_lab/binary_sensor.py
@@ -40,9 +40,11 @@ async def async_setup_entry(
 
     for sensor in AMS_BINARY_SENSORS:
         for index in coordinator.get_model().ams.data.keys():
-            if coordinator.get_model().ams.data[index] is not None:
-                if sensor.exists_fn(coordinator, index):
-                    async_add_entities([BambuLabAMSBinarySensor(coordinator, sensor, index)])
+            ams_instance = coordinator.get_model().ams.data[index]
+            if ams_instance is None or not ams_instance.serial:
+                continue
+            if sensor.exists_fn(coordinator, index):
+                async_add_entities([BambuLabAMSBinarySensor(coordinator, sensor, index)])
 
     for sensor in VIRTUAL_TRAY_BINARY_SENSORS:    
         if sensor.exists_fn(coordinator):
--- a/custom_components/bambu_lab/coordinator.py
+++ b/custom_components/bambu_lab/coordinator.py
@@ -875,10 +875,18 @@ class BambuDataUpdateCoordinator(DataUpdateCoordinator):
         for device in dev_reg.devices.values():
             if config_entry_id in device.config_entries:
                 # This device is associated with this printer.
-                if device.model == 'AMS' or device.model == 'AMS Lite' or device.model == 'AMS 2 Pro' or device.model == 'AMS HT':
-                    # And it's an AMS device
+                is_known_ams = device.model in ('AMS', 'AMS Lite', 'AMS 2 Pro', 'AMS HT')
+                is_placeholder_ams = (
+                    device.model == 'Unknown'
+                    and (DOMAIN, "") in device.identifiers
+                    and (device.name or "").startswith(f"{self.config_entry.data['device_type']}_{self.config_entry.data['serial']}_AMS_")
+                )
+                if is_known_ams or is_placeholder_ams:
+                    # push_status can create an AMS placeholder before version metadata is
+                    # available. Older versions registered that placeholder with an empty
+                    # identifier; always discard it so the real serial can own the device.
                     ams_serial = list(device.identifiers)[0][1]
-                    if ams_serial not in existing_ams_devices:
+                    if is_placeholder_ams or ams_serial not in existing_ams_devices:
                         LOGGER.debug(f"Found stale attached AMS with serial {ams_serial}")
                         ams_devices_to_remove.append(device.id)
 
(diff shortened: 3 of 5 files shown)


AdrianGarside:
Did you capture a repro with logs of the data payload that confirm the initial 'full' status was definitely missing the AMS data?

custom_components/bambu_lab/sensor.py
```diff
@@ -45,6 +45,13 @@ async def async_setup_entry(
 
     for sensor in AMS_SENSORS:
         for index in coordinator.get_model().ams.data.keys():
+            ams_instance = coordinator.get_model().ams.data[index]
+            # push_status can arrive before complete get_version metadata and create
+            # a placeholder AMS with no serial. Device/entity identity depends on the
+            # real AMS serial, so wait for version metadata instead of registering a
+            # transient placeholder that HA cannot later reconcile cleanly.
```
AdrianGarside, line 52: We don't create the devices until we've seen both the status and version data so this explanation doesn't seem right. We don't reach here until both have been observed.

AdrianGarside (commented):
The change generally seems good but some of the explanation for why it's needed doesn't match the existing behavior of waiting until both data and version info has been received before creating the AMS sensors and (as a side effect), the device.

Ali229:
Good point. I didn't capture the original startup MQTT sequence, so I can't confirm that the initial full status or `get_version` was missing AMS data. That part of my explanation was an inference from the resulting `Unknown`/empty-identifier AMS device, so I should correct that.

I still think the PR is addressing the right recovery problem. What I no longer want to claim without a captured repro is the exact startup ordering that causes the stale registry state.

image


I still have the affected HA instance in the broken state and have intentionally not reloaded the integration. I checked the current diagnostics and this is what it shows right now:

* `get_version` contains both `ams/0` and `ams/1`, each with valid AMS08 hardware info, firmware version, and serial.
* The stored `push_status` contains AMS IDs 0 and 1 with all 8 trays populated.
* `device_state.ams` also contains both AMS instances with `model = "AMS"`, valid serials, humidity/temperature data, and tray data.
* Despite that, Home Assistant currently has only one AMS device registered for this printer. It shows `Unknown`, has an empty Bambu identifier, and has zero entities.
* The HA device page also shows the AMS from my other X1C under "Linked devices", which may be related to the empty identifier.

So in the currently broken instance, pybambu has already recovered the correct AMS state internally, but HA has not reconciled the device/entity registry afterward.

The PR's reinitialization and stale-placeholder cleanup are intended to address that recovery gap. I am happy to adjust the exact trigger or lifecycle point if there is a better place in the existing flow to detect and reconcile this state.

Manually reloading the integration normally restores the AMS entities, but I have left this instance broken for now so we can inspect it further if useful.

I also checked the current HA logs and there are no relevant `bambu_lab` errors. I do not have debug logging from the original transition into this state, but I can capture the full sequence next time it reproduces if that would help.

AdrianGarside:
"The HA device page also shows the AMS from my other X1C under "Linked devices", which may be related to the empty identifier."

That seems like it could be related to this issue:
https://github.com/greghesp/ha-bambulab/issues/2089

My theory for that one is that initial 'full' status is missing the AMS data (i.e. a firmware bug). And I think that would be consistent with what you've observed?

Ali229:
I dug into #2089 and compared it against the actual changes in this PR. I think they are related, but I don't think they are exactly the same bug, and #2113 does not directly fix the stale-removal path described in #2089.

#2089 is about `_printer_ready()` removing an already-valid AMS device because `ams.data` is incomplete when the stale-device cleanup runs. This PR does not currently change that removal logic.

What I can now confirm on my instance is slightly different. I just powered on my other X1C and its real AMS device and entities appeared automatically, without reloading HA or the integration. The originally affected X1C has also recovered its two real AMS devices.

However, both printers still have the old `Unknown` AMS device with an empty identifier and zero entities alongside the correct AMS devices.

So the concrete issue I can reproduce for #2113 is that an incomplete AMS identity can leave behind an `Unknown`/empty-identifier placeholder, and that placeholder is not cleaned up when the proper AMS devices are later created. The placeholder cleanup in this PR directly addresses that.

Your theory about an incomplete initial payload could still explain how that placeholder gets created, and it may be the same upstream firmware behavior that triggers #2089, but I don't have the original bad payload from my instance to prove that connection.

For #2089 specifically, it looks like we would still need to make the stale AMS removal safer when the current AMS list may be incomplete. I don't think I should claim this PR fixes that path as written.

Would you prefer that I keep #2113 scoped to the late-identity/placeholder recovery, or extend it to also guard the stale AMS removal from #2089?

Ali229:
Got a clean reproduction today with the extra logging enabled, and I think this confirms the failure path pretty clearly.

### Printer One

On startup, the first `get_version` had **no AMS module at all**.

The first full `push_status` also reported no AMS:

- `ams` list was empty
- `ams_exist_bits = 0`
- tray bits were `0`

When `_printer_ready()` ran, the integration's AMS state was `{}`.

A little later, without reloading anything, pybambu had recovered the real AMS internally with the correct model, serial, tray data, humidity, and temperature. But Home Assistant still had **no AMS entities for Printer One**.

### Printer Two

Its first `get_version` only contained **AMS 0**. AMS 1 was missing.

The first full `push_status` had:

- `ams_exist_bits = 3`, so the printer knew two AMS units existed
- `tray_exist_bits = ff`, so all 8 trays existed
- but the actual `ams` array was still empty

When `_printer_ready()` ran, pybambu only had AMS 0.

Later, again without a reload, pybambu recovered internally and had both AMS 0 and AMS 1 with all tray data. Home Assistant still only had the entities for AMS 0. AMS 1 never got created.

So the sequence I can now reproduce is:

1. The printer initially sends incomplete AMS data.
2. `_printer_ready()` runs with that incomplete AMS state.
3. HA creates only the AMS entities it knows about at that moment, or none at all.
4. Later messages fill in the correct AMS state inside pybambu.
5. HA does not rerun the device/entity setup, so the missing AMS stays missing until a reload/restart.

This looks very close to the behavior discussed in #2089, and it also explains the issue I originally opened this PR for.

Based on this reproduction, I think the PR is targeting the right recovery path: when the missing AMS identity arrives later, retrigger the existing reinitialization so HA can create the missing devices/entities.

I don't think this PR directly fixes every part of #2089 though, especially the stale-device removal logic in `_printer_ready()`.

The next useful test would be to reproduce this same startup condition with the PR applied and confirm that Printer One gets its AMS later and Printer Two gets AMS 1 without needing a reload.

AdrianGarside:
Excellent! Then I think your changes will fix the issue you were seeing and also I expect will largely fix #2089. The device will still get removed but should now get re-added when the printer finally gets around to saying it has it. I'm not seeing a need to try and mitigate the delay since the stale device removal is a valid scenario when you move AMSs between printers but if we see a scenario where the removal and delayed re-addition is a real problem we can revisit that.

